PR #285, opened 2014-12-14
Add "supported" tag for maps

KingRaptor:
Now using its own DB column


Licho1:
Oops i forgot to merge it, it has to be changed for C# code first now, i can help to guide - ideal mini tutorial

Licho1:
To reimplement this DB change do this:

1) open package manager window in VS http://i.imgur.com/MYiQTcg.png 
2) in the window set "default project" to "ZkData"
3) "update-database" into the package manager window, this will create your local db structure to match the servers (this is only needed first time)
4) modify the resource class - simply add the MapIsSupported c# property 
5) "add-migration SomeName" into the package manager window will create new migration class - review that and if happy commit

KingRaptor:
It turns out all the needed DB changes were already done during the migration, reopening and pulling.

knorker:
In past it was often discussed that just a boolean supported-tag would not be enough.
If multiple mods are supposed to be playable (which I understood was still a goal) then a way to indicate *which mod* the map supports is nessecary. 
Some of those threads are linked in this issue, too. 

Either that the lobby's map-selection reads the mod's https://springrts.com/wiki/Validmaps.lua or by having a supported-tag that can be set seperately for different games.

This also does not solve the problem of "silly" maps:
For zero-K a map might be "silly" but for another game it might be normal.

KingRaptor:
> If multiple mods are supposed to be playable (which I understood was still a goal)

I can only speak for myself, but with the way server split went this is probably no longer the case. At least there is unlikely to be any particular effort to guarantee support.

PR #288, opened 2014-12-15
Update HtmlHelperExtensions.cs

mojjj:
>>Forum / BBCode<<
minor fix for vid tags broken with 3gp files
added max-width for img and vid ( scale to 100% width if larger)
added img to display fullscreen img in new tab/window if possible
added url=... link to use new tab/window if possible (not woring for native links


KingRaptor:
Okay, I think all the changes are in master now.

I could still try pressing the merge button but I'm afraid it'll explode...

PR #319, opened 2015-01-03
Springie comsharing

sprunk:
If comsharing modoption is enabled, Springie will make the teams comshared.

### DIFF
--- a/Springie/Springie/autohost/AutoHost_commands.cs
+++ b/Springie/Springie/autohost/AutoHost_commands.cs
@@ -1193,11 +1193,16 @@ namespace Springie.autohost
             if (!string.IsNullOrEmpty(balance.Message)) SayBattle(balance.Message, false);
             if (balance.Players != null && balance.Players.Length > 0)
             {
+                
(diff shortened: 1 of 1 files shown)


Licho1:
this wont really work because ZKL forces unique teamid...

sprunk:
This is for the autobalance that happens just before start, ZKL wont have time to change before start (and ZKL can be changed to allow comshare too)

Licho1:
imo its a bad place to do it, we are moving away from lobby defined teams etc, there is also server balance which transmits ful llist of ally/ids it can be done there at will...

Licho1:
I would rather merge a complete solution rather than half of support for non zkl lobbies...

sprunk:
It happens in Springie's balance so it doesn't matter what lobby the players use.
It also uses the server list of ally/IDs (it reads from it and uses allyID for teamID).

Licho1:
This is not the place though... there is another one in spring.cs

Licho1:
Also should it really be per entire alliance? Perhaps it can be clan based or something like that.. dunno..

sprunk:
This is mostly for tourneys and super-competitive settings so it can be per entire alliance for now and if it is wanted it can later be extended to multiple levels using the modoption (eg. none/clans/everyone).

Licho1:
ok

PR #350, opened 2015-01-08
corners: make forum more visible

Anarchid:
This deals with the fact that white on black is like graser to the eyeball


Anarchid:
Now also fixes positioning of the layout switch button.

PR #357, opened 2015-01-10
Hide "deleted" accounts

KingRaptor:
Redacts username (and forum avatar) wherever referenced in site outside of user page; hides user page from non-admins


sprunk:
Don't people usually want account deletion so that their passwords etc. don't stay in an insecure database? (applying isDeleted doesnt achieve that)

Licho1:
database will be secured soon anyway

KingRaptor:
Data that doesn't interact with anything else (like passwords, but not match history or such) could always be deleted separately in DB; this is just to remove info on deleted accounts from public visibility.

(Although in the password case, if the person keeps signing in to Spring server it'll probably just be added again; this PR should be changed to fix that)

PR #371, opened 2015-01-15
ZKL: tweak title for Warning bar (don't write "Warning" for every message).

xponen:
please comment if you think the title could use better word


Licho1:
Its ugly to set it from outside the bar. Why not have a separate bar type without the warning?

xponen:
erm, I think it might be ugly too if I make many bar with just small difference in title. 

if the WarningBar's constructor can accept new title, will this be less ugly? I did not thought about this before.

Licho1:
Yes you can set it on constructor/method that adds it thats cleaner than setting it post-hoc after its created imo

xponen:
I used reset and made it like I never commit mistake :)

PR #392, opened 2015-01-17
Added support for multiple branches?

ashdnazg:
My C# is rusty at best, this may not even compile?

### DIFF
--- a/Shared/PlasmaDownloader/EngineDownload.cs
+++ b/Shared/PlasmaDownloader/EngineDownload.cs
@@ -24,8 +24,22 @@ namespace PlasmaDownloader
 
         public static List<string> GetEngineList() {
             var engineDownloadPath = GlobalConst.EngineDownloadPath;
-            var data = new WebClient().DownloadString(string.Format("{0}buildbot/default/master/", engineDownloadPath));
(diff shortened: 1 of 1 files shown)


Shared/PlasmaDownloader/EngineDownload.cs
```diff
@@ -24,8 +24,22 @@ public class EngineDownload: Download
...
-            var data = new WebClient().DownloadString(string.Format("{0}buildbot/default/master/", engineDownloadPath));
-            data += new WebClient().DownloadString(string.Format("{0}buildbot/default/develop/", engineDownloadPath));
+            var branchData = new WebClient().DownloadString(string.Format("{0}buildbot/default/", EngineDownloadPath));
+            
+            var branches = Regex.Matches(branchData,
+                              "<img src=\"/icons/folder.gif\" alt=\"\\[DIR\\]\"></td><td><a href=\"([^\"]+)/\">\\1/</a>",
+                              RegexOptions.IgnoreCase).OfType<Match>().Select(x => x.Groups[1].Value).OrderBy(x => x, comparer).ToList();
+                              
+            bool first = true;
+            string data;
```
Licho1: data = ""
and then you dont need the first check in the cycle below and you can append freely

PR #396, opened 2015-01-17
forum: donator star link to contib page

mojjj:
as per request from 
http://zero-k.info/Forum/Thread/11435

link goes to
http://zero-k.info/Forum/Thread/867


Anarchid:
should go to zero-k.info/Contributions, which is a page rather than a forum topic

mojjj:
i agree. BUT. compare the pages and look which one has better infos to donate, which is why i used this.

mojjj:
meaning contrib pages is so 2013 and incomplete

Anarchid:
The prudent solution is then to upgrade the page.

mojjj:
#397

KingRaptor:
Contrib pages may be 2013 but forum thread is 2012.

Anyway, pulling.

PR #400, opened 2015-01-18
Revert "forum: donator star link to contib page"

Licho1:
Reverts ZeroK-RTS/Zero-K-Infrastructure#396


Licho1:
Absolute URLs are absolutely forbidden now! Use relative paths or globalconst variables, never use http://zero-k.info etc directly as it would break in local and test mode.

PR #401, opened 2015-01-18
upgrade low level socket code to use await/async - might fix #399

### LINKED ISSUE #399: Manually opened hosts can fail to appear
Sometimes springies fail to appear when you want to open a new battle. Click "Open a new battle" and nothing happens, you just see a bar at the bottom with "Waiting for room" written on it. This is an intermittent bug as sometimes the hosts start successfully.

This is quite important, at least in the current setup, because new players may try to start a game but think the server is down then leave.

Licho1:


Licho1:
Connection.cs is now much simpler. I had to add some "microsoft.bcl.async" libs to get await/async support on windows xp, because windows xp does not allow .net framework 4.5 but only 4.0

PR #406, opened 2015-01-19
ZK-info - updated Contributions page - Licho TODO

mojjj:
Licho TODO
add costs summary for 2014
update server details if changed


Licho1:
It's not completely correct, euro symbols dont display correctly, use html entity for them i think its something like &euro;
http://test.zero-k.info/Contributions

PR #409, opened 2015-01-19
Update ContributionsIndex.cshtml

mojjj:


mojjj:
€-symbols

Licho1:
to be corect html entity it should end with ; afaik

PR #410, opened 2015-01-19
Update ContributionsIndex.cshtml

mojjj:
&euro;


mojjj:
i mean fix for the euro sign html representation as proposed by mighty licho.

PR #412, opened 2015-01-20
move js and css from SiteLayout to bundles, add System.Web.Optimization

psimyn:
bundles could probably do with better grouping and naming, then was just a direct copy/paste from the View.


Licho1:
breaks users tooltip, reverting

psimyn:
nicetitle.css was included twice :(

PR #417, opened 2015-01-21
Fix campaign events with a hammer

KingRaptor:
I have no idea if this works but it can't break any worse than what we currently have...


Licho1:
im merging but its unlikely to fix,we need to redo FK references between tables, define them more clearly

PR #420, opened 2015-01-22
fix tooltips, update image urls

psimyn:
fixed 08f3957a1ce6f6de54b9a33faa7f57c3438f225b 
- tooltip styles now work properly (nicetitle.css was duplicated)
- updated all(?) images in css to absolute paths.

Only css missing from there is the ref to branding.css, since it doesn't appear to be used anywhere. If needed it can be readded to siteLayout outside of the bundle

### DIFF
--- a/Zero-K.info/App_Start/BundleConfig.cs
+++ b/Zero-K.info/App_Start/BundleConfig.cs
@@ -1,0 +1,33 @@
+﻿using System.Web.Optimization;
+
+public class BundleConfig
+{
+    public static void RegisterBundles(BundleCollection bundles)
+    {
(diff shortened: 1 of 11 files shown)


Zero-K.info/App_Start/BundleConfig.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Web.Optimization;
+
+public class BundleConfig
+{
+    public static void RegisterBundles(BundleCollection bundles)
+    {
+        bundles.Add(new ScriptBundle("~/bundles/main").Include(
```
Licho1: Maybe we could use IncludeDirectory instead to simplify stuff?

PR #434, opened 2015-01-28
Use github as source for award images

Anarchid:
I'm not actually sure this will work as-is but it seems to when i hack the image src attribute in firebug and we have a test environment yay!

This is somewhat of a requirement for tourney award images but i guess i could svn commit them as well.


Licho1:
WTF?? Seriously??
Why do you do this, simply add teamcity rule to copy necessary files to site on build (and generate unitguide)
Thats the way you do it..
Not this, this is from time before continuous integration...

Anarchid:
Will probably look into teamcity solution after tournament.

PR #435, opened 2015-01-29
Uberserver final murder

Licho1:
:no_good: do not merge yet, WIP


xponen:
Licho, please don't need to remove colour from Startscript, its used by ZKL Skirmisher to run all mods

Licho1:
Start script generator should be rewritten and abstracted so that it does not use "lobby" structures directly. Then skirmish can use color or whatever it wants.
Generator should be in one place and lobby or skirmisher should simply fill its structure(s) and let it make script for them.

PR #467, opened 2015-02-08
ZKL: small fix for Skirmisher and playerlist

xponen:
This playerlist fix isn't related to  "refresh spam" or "can't right click" issue, its just other minor clicking/tooltip bug.
  - if cursor move over empty space below the list of player, it won't always show tooltip of the last player in the list anymore 

Skirmish Tab:
  - move my LUA table reader code into new file, trying to use it with other Form later.
  - fix rapid clicking in Skirmisher playerlist not reliable/not-consistent/clumsy
  - auto set Engine & Game to default version at first start

### DIFF
--- a/ZeroKLobby/MicroLobby/ExtrasTab/SkirmishControl.cs
+++ b/ZeroKLobby/MicroLobby/ExtrasTab/SkirmishControl.cs
@@ -139,8 +139,18 @@ namespace ZeroKLobby.MicroLobby.ExtrasTab
             Setup_ComboBox();
             if (!string.IsNullOrEmpty(Program.Conf.SkirmisherEngine))
                 engine_comboBox.SelectedItem = Program.Conf.SkirmisherEngine;
+            else 
+                engine_comboBox.SelectedItem = GlobalConst.DefaultEngineOverride ?? Program.TasClient.ServerSpringVersion;
+            
             if (!string.IsNullOrEmpty(Program.Conf.SkirmisherGame))
                 game_comboBox.SelectedItem = Program.Conf.SkirmisherGame;
+            else
+            {
+                var gameVer = Program.Downloader.PackageDownloader.GetByTag(KnownGames.GetDefaultGame().RapidTag);
+                if (gameVer!=null)
+                    game_comboBox.SelectedItem = gameVer.InternalName;
+            }
+            
             if (!string.IsNullOrEmpty(Program.Conf.SkirmisherMap))
                 map_comboBox.SelectedItem = Program.Conf.SkirmisherMap;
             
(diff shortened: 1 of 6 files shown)


Licho1:
Why do you have own luatable reader? There are already 2 more in code..
and there are existing solutions on nuget now..

Licho1:
Your code makes no sense, why lua table reader? To read modoptions? Those are available from plasma and are already in metadata cache for every mod already downloaded...

If you get them from unitsync then unitsync does return data processed and no need to process table..

xponen:
It is used to read .sdd file Licho1. I didn't find anything in metadata cache to read ssdd files. 

So far I'm able to read .sdd mod, but I can't read .sdd maps, because I don't know how to generate minimap without UnitSync.

Licho1:
You dont need to read anything! Just use metadatacache.getmod ! and it
contains modoptions
what else you need?  Same with map it gives you minimap

2015-02-09 12:39 GMT+01:00 xponen <notifications@github.com>:

xponen:
Licho1, it can't read .sdd folder!

Also, it give other a positive effect, such as:
  - it read faster than UnitSync do
  - is compatible with Linux
  - a tiny step toward independence againt Unitsync

Licho1:
You are talking nonsense. How do you think ZKL displays map minimaps? It
uses metadata cache nothing more. It is 100% linux compatible...

Don't use unitsync
Don't use manual read
Use metadatacache ...  metadatacache.getmod and metadatacache.getmap  ..
why is it not enough?

2015-02-09 12:46 GMT+01:00 xponen <notifications@github.com>:

xponen:
Because: 
  - Metadata don't read .sdd ( https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/master/Shared/PlasmaShared/SpringScanner.cs#L30 )
  - Plasmashared.SpringScanner can't use UnitSync in Linux (for current ZKL configuration) because its too hard to tell MONO to tell Linux to load a library in ZK Datadir which is a non-default library location
- .sdd files are user generated, it don't exist in server. If it read from Metadata cache, its wrong because its from server.

Licho1:
What are you talking about how do you think it works on linux?
It does not need any scanner..
Metadatacache maintains info about all installed maps and mods in cache files. So if you need to know minimap for installed map its already in cache...

xponen:
.sdd file are user generated for testing!

Licho1:
ok but if its local we dont need modoptions? You can simply return empty?

2015-02-09 13:02 GMT+01:00 xponen <notifications@github.com>:

xponen:
Thanks, but its needed for testing :( . Lets say I'm tweaking my ZK source and added option: "ImpulseJump eXtreme". With ZKL I will be able to see how it look and activate the option to use it in my local game.

Licho1:
Ok in that case we should have some universal table reader in shared code i
think there are already several.

2015-02-09 13:34 GMT+01:00 xponen <notifications@github.com>:

xponen:
It will take soooome time to add a universal table reader in shared code. Mainly because it need to be pretty!

Licho1:
I thought we already have several and there are libs that do it too ...

2015-02-09 13:50 GMT+01:00 xponen <notifications@github.com>:

PR #480, opened 2015-02-13
ZKL PlayerListBox: fix issue # 441 (miscalculate playerlist size in MONO)

### LINKED ISSUE #441: ZKL: Player list scroll issue on Mono
> [Anarchblah1] when I scroll down in the player list in chat, I can't get all the way down
> [Anarchblah1] some miscalculation in the playerlist size
> [Anarchblah1] oddly enough it seems to get worse when the window is larger
> [Anarchblah1] when I make the window bigger to see more names, I can't scroll down as far and ironicly see fewer names iirc

xponen:
REASON:
PlayerListBox's code didn't obey
https://msdn.microsoft.com/en-us/library/system.windows.forms.listbox.itemheight(v=vs.110).aspx
which says "ItemHeight" must return height of first item in the list when displaying multiple items with variable height (the "DrawOwnerVariable" tag).

Weirdly this omission didn't cause issue in NET (probably in NET the ListBox size is recalculated using value returned from "OnMeasureItem()").

FIX:
Override "ItemHeight" property and return the Height of the first item in the ListBox.
fix #441

### DIFF
--- a/ZeroKLobby/MicroLobby/PlayerListBox.cs
+++ b/ZeroKLobby/MicroLobby/PlayerListBox.cs
@@ -16,6 +16,14 @@ namespace ZeroKLobby.MicroLobby
 	    ObservableCollection<PlayerListItem> realItems;
 	    Timer timer;
 	    public PlayerListItem HoverItem { get; set; }
+        public override int ItemHeight{
+            get {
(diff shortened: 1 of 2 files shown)


Licho1:
good job!

xponen:
thanks! :)

PR #487, opened 2015-02-15
Pull Master into Welcome_screen

xponen:
i added several tiny changes to chat stuff, if we merge earlier we can avoid conflict later.


xponen:
problem with KR's "Merge branch master into mapTags2"

xponen:
ops, mistake in reading :/

PR #489, opened 2015-02-17
enable planet wars for testing

sprunk:


Licho1:
you should also change protocolextension.cs or matchmaking wont work

sprunk:
sure, for now i'll just look at the metagame stuff

PR #493, opened 2015-02-20
Springie: undo adding message to tell player to reconnect to lobby

xponen:
I tested and it sound silly ingame (because its a personal message read by others)


Licho1:
It's not personal, it uses "private" batle chat message!!
Message appears in battle room but only for the client not for others.

Issue #98, opened 2014-09-06
{up,down}vote removal defeats the purpose


sprunk:
+1's purpose is to prevent useless "+1" posts. removing it due to abysmal karma defeats this purpose (see princereaper who keeps spamming "i would +1 you if i could")

PR #502, opened 2015-02-25
remove minimum karma to vote

sprunk:
reasoning:
* causes additional spam of "+1" posts (see #98)
* can be easily circumvented by getting +9001 from a silly gif (eg. http://zero-k.info/Forum/Thread/9847?page=2#108816 )
* is a case of karma being per-user and not per-post (encourages farming etc., same reasoning as why per-user karma isn't shown on the user page)
* extremely toxic cases can be handled with a forum ban instead (eg. http://zero-k.info/users/lobbydetail/paShadoWn )
* is not documented

### DIFF
--- a/Shared/PlasmaShared/GlobalConst.cs
+++ b/Shared/PlasmaShared/GlobalConst.cs
@@ -202,7 +202,7 @@ namespace ZkData
 
         public const int ForumPostsPerPage = 50;
         public const int MinLevelForForumVote = 10;
-        public const int MinNetKarmaToVote = -30;
+        public const int MinNetKarmaToVote = -9999;
         public const int PostVoteHideThreshold = -6;
         public const bool OnlyAdminsSeePostVoters = false;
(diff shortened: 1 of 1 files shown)


Licho1:
Don't think that there is serious problem with this. .. its just a form of punishment for low karma..

Anarchid:
It's not hugely problematic, just undocumented and inconsistent.

Licho1:
Removal removes feature. I don't see any problem with feature not documented for people with -30 karma.. we don't document banning systems either.

sprunk:
It's not really a useful feature for reasons above.

Also bans are a common feature everywhere so everyone knows they exist anyway, unlike a voting block.
They are also shown on the userpage so you can know that you got them and why, but you can't see your karma to know if you can vote (even if you knew about the "feature").

Licho1:
I'm against if other devs disagree they can reopen this.

PR #508, opened 2015-02-27
Merge Master into Stable again

xponen:
Zero-K NET4.0 can't build, it make auto-build stuck, maybe after changing name to Zero-K_NET4.0 it might work?

### DIFF
--- a/Springie/Springie/autohost/AutoHost_commands.cs
+++ b/Springie/Springie/autohost/AutoHost_commands.cs
@@ -1000,8 +1000,8 @@ namespace Springie.autohost
                     Respond(e, "requires key=value format");
                     return ret;
                 }
-                var key = parts[0];
-                var val = parts[1];
+                var key = parts[0].Trim(); //Trim() to make "key = value format" ignore whitespace 
+                var val = parts[1].Trim();
 
                 var found = false;
                 var mod = hostedMod;
--- a/ZeroKLobby/MicroLobby/ToolTabs.cs
+++ b/ZeroKLobby/MicroLobby/ToolTabs.cs
@@ -246,6 +246,9 @@ namespace ZeroKLobby.MicroLobby
 
         private void RemoveTab(string key)
         {   
+            if (!controls.ContainsKey(key))
+                return;
+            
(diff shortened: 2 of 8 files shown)


Licho1:
Please dont store ZKL 4 in master anymore, since we decided its not a maintainable but one time "compile" only change no need to do that..
You would have to port entire new ZKL which is massive change with 1200 changed files..

xponen:
maybe when new ZKL release I can slowly put all the stuff into ZKLNET4?

Licho1:
Whats the point, you can make ZKL 4 at any time by doing 3 simple steps..
change projects to .net 4.0 (does not matter that many wont compile), add async BCL, done

xponen:
is that easy??
wont we have to change the event handler in TasClient, and change Task to TaskEx, and edit reference to async BCL too?

Licho1:
Ok and cahnge Task to TaskEx but I believe it can be done in 5 mins or so when it's trained a bit. So I think it will be less wor than maintaining entire folder like this.

xponen:
idk... at moment is still okay for me IMO,

My step is:
I copy paste original folder
  ZeroKLobby, PlasmaShared, LobbyClient, MonoTorrent, and PlasmaDownloader 
into ZeroKLobby_NET4.0, then overwrite all files, then using what I remember for NET4.0 conversion steps, to check file diff 1-by-1 using Tortoise Commit window to avoid changing the NET4.0 stuff.

Then if ZKL changes, I would replace ZeroKLobby_NET4.0.csproj with original ZeroKLobby.csproj and carefully check file diff using Tortoise again.

Maybe at moment is okay, if it get too hard, then I surrender...

Licho1:
Yeah but try merging this into welcome_screen :-)
Thats what i have to do now

xponen:
ooo noo...

xponen:
let me help, I will spend time resolve conflict 1 by 1, and readd NET4.0 so I won't need to copy paste again later ...
pls

xponen:
maybe I include new ZKL into NET4.0 at same time!

PR #515, opened 2015-03-01
remove useless/trolling commands

sprunk:
removes commands not wanted on public hosts:
* !voteboss
* !voterehost
* !votemove

private rooms already boss the creator so rehost/move are still available.

!id removed (springie overrides id)
!saveboxes rights changed to 4 (above boss)


Licho1:
Imo votemove is useful and should be kept. Other commands like voteboss and voterehost also have their use and so far I have only seen legitimate use.
(Voteboss to arrange custom/chicken teams, voterehost to fix broken host/reset to defaults).

id and saveboxes changes are fine.

When you remove command please also remove the actual implementation (ComId) and case "id" section in autohost class.

Anarchid:
>  (Voteboss to arrange custom/chicken teams, voterehost to fix broken host/reset to defaults).

Chickens custom teams can be achieved with !voteforcestart, however. 

Voterehost could probably stay unless Sprung demonstrates an exploitative mode (which is not unlikely).

sprunk:
* removed implementations and cases in autohost class for the removed commands
* removed !admins (doesn't work, a working admin list is on the site already)
* brought back !votemove
* !voterehost was actually already removed by KR in https://github.com/ZeroK-RTS/Zero-K-Infrastructure/commit/e18a71e9600723a67d638d68ea34777041c35e04 so i kept status quo
* !voteboss stays removed because it allows access to admin commands like !setmaxplayers which allow breaking of public rooms, while everything else that it is needed for (like custom settings) can be done through normal votes (eg. !votesetoptions and !voteforcestart)

PR #517, opened 2015-03-02
Show Elo on own user page

sprunk:


Skasi:
Remove '<class="admin">' in line 136, it's a red dashed border used to mark admin-only things.

xponen:
also, might need to add some text saying this is only visible to user.
If not I worry user will worry too much about people judging them thru elo.

Anarchid:
They won't be able to see it on other players' pages, so this should be evident.

Licho1:
Dislike this. Elo is not meant for "ranking" it's only meant to be hidden technical value for team balancing.

Skasi:
> Elo is not meant for "ranking"
http://zero-k.info/Ladders

Princereaper:
Why hav a ladder. I would like to know my elo. What wrong with me knowing my own elo???

PR #518, opened 2015-03-02
obediation election deprecation notification

sprunk:


Licho1:
Voteboss is not meant to be deprecated.

Anarchid:
1) There are no useful use cases for !voteboss in public hosts.
1.1) Exception is using !voteboss to revert damage done by a malicious boss earlier

2) The custom hosts automatically receive a boss. If this boss deigns to remove himself, this does not require a !voteboss. If other people barge into a private game (say, by abusing !move), !voteboss can be used to take the room over.

3) Your revert now made it fail silently ("i don't know command !voteboss"). Morale of story: don't reflexively click the button.

Licho1:
There are uses on public host when you want to setup custom game without interference of others or custom balance.

Licho1:
Another use is when boss leaves from bossed custom host.

Anarchid:
> Another use is when boss leaves from bossed custom host.
Can unboss themselves before doing so. 

> There are uses on public host when you want to setup custom game without interference of others or custom balance. 

Can use custom room for that.

Licho1:
Yes but what happens with custom room when boss leaves? Isn't it simpler to disable voteboss on team hosts if it causes troubles there?

Anarchid:
Disabling voteboss on pub hosts is an acceptable compromise (removes admin work for me).

It still would allow overtaking custom games, but i guess those can be password-protected once !move is fixed to respect passwords.

Licho1:
I thought more about it ok remove voteboss completely. There will be different solution for "custom hosts". I would like mode where self-changes are blocked. Where you cannot change your team.

PR #519, opened 2015-03-02
attempt to improve git notifications

Anarchid:
- adds pull request notification
- adds repository identification
- probably doesn't work 
- i have no idea why [{0}] was formerly completely abjectly absent from the message

### DIFF
--- a/Zero-K.info/Controllers/GithubController.cs
+++ b/Zero-K.info/Controllers/GithubController.cs
@@ -31,15 +31,17 @@ namespace ZeroKWeb.Controllers
             dynamic payload = JObject.Parse(Encoding.UTF8.GetString(data));
 
             string text = null;
+            
+            Object[] values;
 
             switch (eventType) {
                 case "issues":
-                    var values = new [] {payload.repository.name ,payload.sender.login,  payload.action, payload.issue.title, payload.issue.html_url};
+                    values = new [] {payload.repository.name ,payload.sender.login,  payload.action, payload.issue.title, payload.issue.html_url};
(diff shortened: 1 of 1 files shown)


Zero-K.info/Controllers/GithubController.cs
```diff
@@ -31,10 +31,18 @@ public ActionResult Hook()
             dynamic payload = JObject.Parse(Encoding.UTF8.GetString(data));
 
             string text = null;
+            
+            Object[] values;
 
             switch (eventType) {
                 case "issues":
-                    text = string.Format("{0} has {1} issue {2} {3}",payload.sender.login,  payload.action, payload.issue.title, payload.issue.url);
+                    values = new [] {payload.repository.name ,payload.sender.login,  payload.action, payload.issue.title, payload.issue.html_url};
```
Licho1: Why did you use "values" field? Makes no sense, it could be passed directly to string.format as extra params.

Anarchid: The line was super long with both string.Format and values list. My religion forbids lines of such length, so i sacrificed a a LOC to appease the gods of readability.

Licho1: Modern editors have auto wrap and line can be separated, you can split line so that for example each argument is on separate line.

Anarchid:
Fair enough. I don't really insist on this bit.

Zero-K.info/Controllers/GithubController.cs
Anarchid: Fair enough.

Issue #243, opened 2014-11-22
Springie !setengine bugged / password protected games setengine


ArchShaman:
Exploit reports setengine is bugged for them (running zkl on linux).

Secondly, password protected games with setengine does not auto-rejoin the player.

blah1:
In case nobody realized it yet, this is caused by Springiee using a Global variable for "CurrentEngine" (or whatever it's called). Steps to verify:
1. PM to any Springiee host: !spawn title=Room A,mod=zk:stable
2. Enter Room A and give the command: !setengine 88.0 (or whatever you want)
3. Springiee## replies with "[3:33 PM]  * Springiee6 Preparing engine change to 88.0
[3:33 PM]  * Springiee6 rehosting to engine version 88.0" and rehosts the room
(you can rejoin the room, I didn't bother for this test)
4. PM to any Springiee host: !spawn title=Room B,mod=zk:stable
5. Enter Room B and give the same command as before: !setengine 88.0
6. Springiee## replies with "[3:34 PM]  * Springiee14 Preparing engine change to 88.0" and doesn't change the engine

The bug is that the engine in use is shared between hosts, so if host A changes to 91.0, host B thinks that it already has 91.0 even though it didn't change

xponen:
nice theory! 
i'm checking Springie code in case I can see this global var

PR #522, opened 2015-03-03
Springie: fix buggy !setengine issue

xponen:
Setengine normally occur when a global "SpringPaths" call a "SpringVersionChanged" event to all Autohost at same time. This work perfectly fine if we wanted to change engine for all Autohost at same time, but !setengine is coded to effect only a single Autohost. As result since "SpringVersionChanged" was called only once per !setengine, it won't get called again when commanded in other Autohost.

The fix is to tell each Autohost to use its own "SpringVersionChanged" event from local "SpringPaths" (already present in the code). This immediately fix https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/243

Have done some test.

### DIFF
--- a/Shared/PlasmaDownloader/PlasmaDownloader.cs
+++ b/Shared/PlasmaDownloader/PlasmaDownloader.cs
@@ -68,18 +68,25 @@ namespace PlasmaDownloader
             packageDownloader.Dispose();
         }
 
-        public Download GetAndSwitchEngine(string version) {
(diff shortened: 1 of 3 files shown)


Licho1:
It's ok solution.
Eventually I think its best to remove all these "downloadandswitchengine' and instead let "spring.cs" auto DL and wait for resources its missing - engine, mod, map.

Issue #529, opened 2015-03-05
ZKL intermittently starts DefaultOverride engine instead of room specified engine


Anarchid:
Probably tied to server version somehow. Some chat reports:

```
16:53:37	Svatopluk	now it's starting spring 91
16:53:47		+ now crash
16:54:18		Anarchid infolog.txt useless, it just started wrong spring version
16:54:23	Anarchid	wow that's a serious bug
16:54:45	Svatopluk	now rejoin room + rejoin game = starting correct, 98 spring
```

```
16:55:33	ivand	sometimes it tries to start version of engine you !setengine'd prior to the recent !setengine
16:55:53		like you !setengine 88.0 firstly
16:56:00		then !setengine 91.0 next
```

xponen:
might be problem with ZKL in XP
I just got ZKL-log from Huj, trillion of errors with Async stuff, will have to read slowly
rghost.net/7h5jHc2PC

Anarchid:
This is unlikely to be an XP issue since i'm pretty sure ivand and svatopluk have newer versions.

Additionally, this issue is probably solved and i'm only keeping it open for a couple more days just in case. There have been no recurrent reports of 91.0 being launched since the default version of everything was set to 98.0.1-398 in GlobalConst.

Anarchid:
Hmmm actually if this happens again, it can be reopened. One day is probably enough data to suggest it being fixed, since it was super prominent when it was occuring.

Anarchid:
This still happens, as engine version progresses with new awesome fixes and bugs.

Consequently, people complain about desyncs when they shouldn't.

xponen:
will keep an eye on subtle trigger, this will take some time.

Licho1:
dont bother with subtle trigger, its destined for full rewrite, let me know if you are interested i can explain in chat

xponen:
found it resetting to default in Skirmish tab, is investigating

xponen:
probably only happen to ZKL net4.0
fixed and copy+pasted new build at zero-k.info/www/lobby as  ZKL ver 1.3.2.62 (net4.0)

PR #531, opened 2015-03-05
GlobalConst: make 98.0.1-398-g80e6cfa default to try to workaround #529

Anarchid:


Licho1:
This has massive consequences. Make 100% sure it works even for new users - that for example ZKL starts mission on 98 properly

Licho1:
(With nothing preinstalled)

Anarchid:
Hm. Will test more, revert if breaks.

Anarchid:
Missions seemed to work on a machine completely purged out of anything ZK.

I think the installer is not automatically updated, because it tried to dowload 91.0 initially and also immediately demanded an update.

Licho1:
Installer contains no mod or mod download.

Anarchid:
Hmm, right, it did download zero-k.exe before installing it. Does that mean that there's another obscure global somewhere telling it to dl 91?

Licho1:
It might be in the mission or mission script i dont know..

2015-03-06 11:03 GMT+01:00 Анархид <notifications@github.com>:

Issue #521, opened 2015-03-02
lobby is incompatible with mono?


jhoblitt:
I'm unable to get the lobby to function with the version of mono shipped with Fedora 21:

```
[master] ~/games $ rpm -qa | grep mono
mono-addins-0.6.2-11.fc21.x86_64
mono-extras-2.10.8-8.fc21.x86_64
mono-basic-devel-2.10-7.fc20.x86_64
mono-zeroconf-0.9.0-13.fc21.x86_64
monodoc-2.10.8-8.fc21.x86_64
mono-yui-1.1.0-3.fc21.x86_64
mono-locale-extras-2.10.8-8.fc21.x86_64
mono-tools-ilcontrast-2.10-12.fc21.x86_64
wine-mono-4.5.4-1.fc21.noarch
mono-shogun-3.2.0.1-0.27.git20140804.96f3cf3.fc21.x86_64
mono-core-2.10.8-8.fc21.x86_64
mono-winforms-2.10.8-8.fc21.x86_64
liberation-mono-fonts-1.07.4-4.fc21.noarch
mono-reflection-0.1-0.9.20110613git304d1d.fc21.x86_64
mono-web-devel-2.10.8-8.fc21.x86_64
mono-debugger-2.10-9.fc21.x86_64
mono-nunit-devel-2.10.8-8.fc21.x86_64
mono-web-2.10.8-8.fc21.x86_64
mono-icon-theme-14.12.1-1.fc21.noarch
mono-reflection-devel-0.1-0.9.20110613git304d1d.fc21.x86_64
mono-bouncycastle-1.7-8.fc21.x86_64
mono-devel-2.10.8-8.fc21.x86_64
mono-tools-monodoc-2.10-12.fc21.x86_64
mono-data-oracle-2.10.8-8.fc21.x86_64
mono-cecil-flowanalysis-0.1-0.22.20110512svn100264.fc21.x86_64
mono-debugger-devel-2.10-9.fc21.x86_64
mono-tools-2.10-12.fc21.x86_64
mono-data-postgresql-2.10.8-8.fc21.x86_64
mono-mvc-2.10.8-8.fc21.x86_64
mono-wcf-2.10.8-8.fc21.x86_64
mono-tools-gendarme-2.10-12.fc21.x86_64
mono-cecil-0.9.5-3.20140924git6d1b7d0.fc21.noarch
mono-basic-2.10-7.fc20.x86_64
[...]

PR #532, opened 2015-03-05
Zkl & ZklNET4.0: won't use "tooltip Always on top" fix in Linux

xponen:
because user32.dll not found exception in Linux Fedora

hopefully fix: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/521

I have uploaded zklNet4.0 version to server

### DIFF
--- a/ZeroKLobby/ToolTips/ToolTipHandler.cs
+++ b/ZeroKLobby/ToolTips/ToolTipHandler.cs
@@ -189,8 +189,11 @@ namespace ZeroKLobby
                         ny = mp.Y - tooltip.Height - 8;
                     }
 
-                    SetWindowPos(tooltip.Handle, HWND_TOPMOST, 0, 0, 0, 0, SWP_ToolTipOption); //refresh tooltip's Z-order to be on top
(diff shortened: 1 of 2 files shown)


ZeroKLobby/ToolTips/ToolTipHandler.cs
```diff
@@ -189,8 +189,11 @@ public class ToolTipHandler: IMessageFilter, IDisposable
                         ny = mp.Y - tooltip.Height - 8;
                     }
 
-                    SetWindowPos(tooltip.Handle, HWND_TOPMOST, 0, 0, 0, 0, SWP_ToolTipOption); //refresh tooltip's Z-order to be on top
-                    tooltip.SetDesktopLocation(nx, ny); //note: SetWindowPos() positioning won't work in Linux (so we shouldn't replace this with SetWindowPos)
+                    if (Environment.OSVersion.Platform != PlatformID.Unix)
+                    {
+                        SetWindowPos(tooltip.Handle, HWND_TOPMOST, 0, 0, 0,0, SWP_ToolTipOption); //refresh tooltip's Z-order to be on top
```
Licho1: You can set it topmost without need to do this call, there shuld be ordinary Topmost property

PR #541, opened 2015-03-10
Zkl & zklNet4: local replay "can't find dev engine fix"

xponen:
accidentally included the word " develop" when reading the name from demo file


xponen:
this is quickfix, merging right away.

PR #542, opened 2015-03-10
merge LocalReplay fix from Master to Welcome_screen too.

xponen:
making sure LocalReplay in new_ZKL can also be tested to run replay from dev engine.


xponen:
this is not so dangerous, I merge right away.

PR #549, opened 2015-03-17
ZKL: fix intermittent default engine issue.

xponen:
should fix https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/529

Is caused by an engine check which always reset engine to default

The origin of bug: 
https://github.com/ZeroK-RTS/Zero-K-Infrastructure/commit/39e1de345d5e41e4f7052133441f45dc32f8cd5b , an irregular behaviour noted in #529 is due to it using "PackageDownloaded" event as delay, which is after ZKL finish downloading Rapid list from server at startup.

### DIFF
--- a/ZeroKLobby/Utils.cs
+++ b/ZeroKLobby/Utils.cs
@@ -304,12 +304,13 @@ namespace ZeroKLobby
         }
 
 
-        public static bool VerifySpringInstalled() {
+        public static bool VerifySpringInstalled(bool showMessageBox = true) {
             if (Program.SpringPaths.SpringVersion == null || !Program.SpringPaths.HasEngineVersion(Program.SpringPaths.SpringVersion)) {
-                MessageBox.Show("Cannot start yet, please wait until engine downloads",
-                                "Engine not prepared yet",
-                                MessageBoxButtons.OK,
-                                MessageBoxIcon.Information);
+                if (showMessageBox)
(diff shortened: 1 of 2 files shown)


ZeroKLobby/Utils.cs
```diff
@@ -304,12 +304,13 @@ public static void SetIeCompatibility()
...
 
 
-        public static bool VerifySpringInstalled() {
+        public static bool VerifySpringInstalled(bool showMessageBox = true) {
             if (Program.SpringPaths.SpringVersion == null || !Program.SpringPaths.HasEngineVersion(Program.SpringPaths.SpringVersion)) {
-                MessageBox.Show("Cannot start yet, please wait until engine downloads",
-                                "Engine not prepared yet",
-                                MessageBoxButtons.OK,
-                                MessageBoxIcon.Information);
+                if (showMessageBox)
```
Licho1: The messagebox should imo be dialog which shows all pending downloads and auto hides when they are done and allows. Best in welcome_Screen branch.

xponen: I agree, its like Firefox's download window.

Will require to create a new form in welcome_Screen where we can put existing download bar into it.

Licho1:
Imo no need to put existing bars in , it can be redesigned too... why it should be a huge bar with minimap? Not needed imo..

xponen:
I;m going to merge this and release new ZKL for this bug

PR #555, opened 2015-03-20
Disallow spectators to use most !commands

sprunk:
Technical and info commands are kept:

!spawn
!adduser
!transmit
!voteresign
!notify
!predict
!springie

### DIFF
--- a/Springie/Springie/autohost/AutoHost.cs
+++ b/Springie/Springie/autohost/AutoHost.cs
@@ -209,7 +209,7 @@ namespace Springie.autohost
             if (tas.MyBattle == null) return true;
             if (spring.IsRunning)  {
                 PlayerTeam user = spring.StartContext.Players.FirstOrDefault(x => x.Name == e.UserName && !x.IsSpectator);
-                return ((user != null) && !user.IsSpectator);
+                return ((user == null) || user.IsSpectator);
             } else {
                 return !tas.MyBattle.Users.Values.Any(x => x.LobbyUser.Name == e.UserName && !x.IsSpectator);
             }
--- a/Springie/Springie/autohost/AutoHost.cs
+++ b/Springie/Springie/autohost/AutoHost.cs
@@ -200,6 +200,11 @@ namespace Springie.autohost
     }*/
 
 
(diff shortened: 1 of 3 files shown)


Licho1:
I think that votemove should also be allowed because it moves spec to another host.
Or !help..
And it was also purpose to have voterehost and votekick available due to extreme circumstances - like all people resigned except one lagger or host broken.

Springie/Springie/autohost/AutoHost.cs
```diff
@@ -233,7 +238,7 @@ public void Start()
                     for (int i = 0; i < c.ListenTo.Length; i++) {
                         if (c.ListenTo[i] == e.Place) {
                             // command is only for nonspecs
-                            if (!c.AllowSpecs) if (tas.MyBattle == null || !tas.MyBattle.Users.Values.Any(x => x.LobbyUser.Name == e.UserName && !x.IsSpectator)) return false;
+                            if (!c.AllowSpecs && !GetUserAdminStatus(e)) if (tas.MyBattle == null || !tas.MyBattle.Users.Values.Any(x => x.LobbyUser.Name == e.UserName && !x.IsSpectator)) return false;
```
Licho1: There are several "levels" of spec.
1) lobby spec which can be changed at any time - you used this
2) ingame spec, which cannot be faked. 

I would suggest using ingame spec if game is running.

sprunk:
Using ingame spec where applicable is a good idea, i will look into it.

Currently everything that is needed can be done through buttons in the lobby so !help isn't much needed (but i'll change to allow it anyway).

Instead of allowing votes to fix extreme circumstances it's better to prevent extreme circumstances from existing. For example if laggers don't get resigned then instead of votekicking them, better to change resign to affect laggers ( see https://github.com/ZeroK-RTS/Zero-K/pull/641 ). Non-specs can also still use the commands to deal with any problems.

sprunk:
Now uses ingame spec if spring is running.
Allowed !help, !ring, and !votemove.

Licho1:
pull master manually to hjere to fix auto merge

sprunk:
#557 for fixed auto merge

PR #556, opened 2015-03-22
votemove: require explicit consent

sprunk:


Springie/Springie/autohost/Polls/VoteMove.cs
```diff
@@ -16,7 +16,7 @@ public class VoteMove: AbstractPoll
 
         protected override bool PerformInit(TasSayEventArgs e, string[] words, out string question, out int winCount)
         {
-            winCount = 0;
+            winCount = (tas.MyBattle != null) ? tas.MyBattle.Users.Count : 1;
```
Licho1: If it moves only those that agree why demanding a majority?

sprunk:
to give everyone chance to vote.

by default votes need nonspecs/2. this means if there are eg. 6 players and 4 specs it would end after 3 votes even though there are 10 people so the remaining 7 people would not be able to vote.

PR #559, opened 2015-03-23
Springie: allow room creator to use !commands from spec

sprunk:


xponen:
this is sprunk's 2nd fix for spec !command

PR #562, opened 2015-03-28
Battle detail: correct times shown

sprunk:
* Duration now shows warmup time in addition to ingame.
* Player lose time is shown in ingame time instead of total time.


sprunk:
on second thought changing DB schema is too dangerous, will resubmit a similar patch that doesnt touch DB soon

Issue #477, opened 2015-02-09
Chronological Reports Log


Anarchid:
Currently reports are filed to #zkadmin and are then logged to individual user pages.
If a report has been consumed by a subscribed lobby it is not very easy to retrieve afterwards.
It would be much easier to administrate things happening during slow hours if there was a persistent, chronological, easily accessible log of reports.

Bonus points if there's a button for "dealt with".

Anarchid:
^

PR #564, opened 2015-03-28
add a report log to site

sprunk:
see #477

### DIFF
--- a/Zero-K.info/Controllers/UsersController.cs
+++ b/Zero-K.info/Controllers/UsersController.cs
@@ -91,6 +91,12 @@ namespace ZeroKWeb.Controllers
             return View("AdminUserDetail", user);
         }
 
+        [Auth(Role = AuthRole.LobbyAdmin | AuthRole.ZkAdmin)]
+        public ActionResult ReportLog ()
+        {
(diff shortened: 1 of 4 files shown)


Zero-K.info/Views/Users/ReportLog.cshtml
```diff
@@ -0,0 +1,18 @@
+@using ZkData
+@using ZeroKWeb
+           
+@{
+    Page.Title = "Report log";
+}
+
+<h2>Reports</h2>
+
```
Licho1: This table is missing <table> </table>, make sure produced html is valid

Zero-K.info/Controllers/UsersController.cs
```diff
@@ -91,6 +91,12 @@ public ActionResult AdminUserDetail(int id)
             return View("AdminUserDetail", user);
         }
 
+        [Auth(Role = AuthRole.LobbyAdmin | AuthRole.ZkAdmin)]
+        public ActionResult ReportLog ()
```
Licho1: Newly created url should be linked from "admin " section of admin's home pages

PR #567, opened 2015-03-30
another fix attempt

sprunk:


Licho1:
You can and should test this stuff *locally* .. it will run in VS2013 community edition - incl. website

PR #572, opened 2015-03-31
remove clan limit

sprunk:


Skasi:
Just providing a bit of content:

[23:02]  Persius bad clan design is negatively impacting this game
[23:03]  Persius I try to join the skasi clan but it say full or
[23:03]  Persius why is it so limited in capacity?
[23:04]  Persius you want the clustermuck, but complain and then you dont support the clan properly
[23:04]  Persius makes no logical sense
[23:06]  Sprung the premise is false
[23:07]  Persius but I want to play as organised faction in clan, and can not
[23:07]  Persius because of strange limit
[23:08]  Persius leading to dis organised chaos in games

TL;DR: New player wants to join my clan for proper teamplay using Mumble, but ZK is being bureaucratic.

Licho1:
Imo do not remove clan limit until reliable friends system exists.
Atm clans are used to balance together, having too big clans defeats that purpose.

Skasi:
Well in that case it's the players fault for having 10+ clan members in one room.

In my clan with the current limit only 0 to 2 members are online at the same time, that's ~10% of the members. My clan could have four times as many members and still wouldn't fill an entire team in the current 10v10 team room meta, thus not defeat its purpose.

PR #573, opened 2015-04-02
Revert "Use proper "away" icon by default (UNTESTED!)"

xponen:
Reverts ZeroK-RTS/Zero-K-Infrastructure#571
Turn out both icon is used in different places. So swapping is not good.
Will redo this by adding transparency to the icon later.


Skasi:
As Sprung said, just use away.png for everything and merge "away" and "away1".

Where is the .bmp version used? I [couldn't find any use case with the search function](https://github.com/ZeroK-RTS/Zero-K-Infrastructure/search?utf8=%E2%9C%93&q=away1), otherwise I would've replaced it myself.

PR #574, opened 2015-04-02
use png instead of bmp for afk icon

sprunk:


xponen:
I added transparency to existing bmp!

Licho1:
Cannot merge atm also the final bakcground is blueish

xponen:
I have pulled a trasparent grey zZ here https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/577 , so no need this anymore

maybe its a bit blurry, will fix

Issue #570, opened 2015-03-31
While using ZKL's player search the player list is not updated


Skasi:
Note how on the left you can see a lot of "xxx joined." messages, but the players don't show up on the right. This happens only when the Search bar on the top right is used. When not searching for any players, the list updates properly.

![zkl playerlist bug](https://cloud.githubusercontent.com/assets/132906/6926288/87b7840c-d7e5-11e4-872a-0ce174d0380d.png)

xponen:
this seem easy to fix, let me do

PR #576, opened 2015-04-03
Zkl: playerlist search fix/ improvement

xponen:
this fix https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/570
and fix it showing wrong result if you search (eg: in ZK channel) while you are in battleroom.

The bug was:
the 1st bug is because it wasn't intended to update while you search, but I don't see any regression of any function or feature while it does, so I made it obey https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/570 
Line https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/576/files#diff-3bb36f6c9dda362536527e8b698c38d1R445 

also, 2nd bug is because the player's SortOrder is also effected by player's team (for battleRoom), so I made it detect this case and skip them while searching
Line https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/576/files#diff-b5d6baab0eb4cbe29263bf07ee6a0a2cR250

additionally I made some magic value to be a global constant (code cleanup) so that its easy for dev to find using IDE
Line: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/576/files#diff-b5d6baab0eb4cbe29263bf07ee6a0a2cR16

### DIFF
--- a/ZeroKLobby/MicroLobby/BattleChatControl.cs
+++ b/ZeroKLobby/MicroLobby/BattleChatControl.cs
@@ -132,20 +132,20 @@ namespace ZeroKLobby.MicroLobby
 			{
 				var missionSlot = GetSlotByTeamID(bot.TeamNumber);
 				newList.Add(new PlayerListItem
-				            { BotBattleStatus = bot, SortCategory = bot.AllyNumber*2 + 1, AllyTeam = bot.AllyNumber, MissionSlot = missionSlot });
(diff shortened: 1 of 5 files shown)


Licho1:
Thank you!
Though eventually we will need proper fix (full rewrite since the code sux and does not support "slots" mode planned)

PR #577, opened 2015-04-03
zkl: replace grey zZ icon with transparent grey zZ

xponen:
this re-implement the reverted https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/571

have tested in normal white background and unusual black background. Work quite okay

zz transparent


xponen:
updated transparency to 
zz transparent 3 in latest commit

Issue #558, opened 2015-03-23
Rapid causes epic hang, fails to download


mtroyka:
When using rapid manually (ie through the settings panel), double clicking any item on the left sometimes/often causes ZKL to freeze for minutes at a time before finally responding, during which time it gets disconnected from the server so that when the download finally starts it never progresses. When rapid is called automatically (ie for map downloads and zk updates) I've never seen this happen that I can recall.

For whatever reason this doesn't seem to happen nearly as badly when ZKL has just been started, which is the only way I got it to download anything at all.

xponen:
could you share the content of your rapid?:
"your-datadir\packages\selected.list"

probably will help in replicating the bug.

Licho1:
It's a known bug i changed the mode from async to sync when rewriting parts of code to use "await", due to some issues.

xponen:
you mean "LoadMasterAndVersions()"?
we should return to Async and because only 1 code call it, we can just use  "do Thread.Sleep(); while(isRefreshed)" to get same behaviour as "await".

Licho1:
Its not so easy, because packages are now not auto refreshed every 2 minutes, due to abma. Instead they are only checked when you request download, and download itself has to follow and function needs to return "Download" object to caller.
So you would have to block anyway..

*BUT* it's that way because its only temporary, the plan is to remove package refresh from client and move it to server (site) and store data in DB .. same stuff as I commented on your previous fix commit.

xponen:
can I help abma more by increasing my spam timer from 3 second to 2 minute?

Licho1:
You do it 3s but only when dl is requested, which should be ok for abma

abma:
its not because of me, its because of https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/198

shaun0x00:
I don't know what you are attempting to fix.
ZKL is still opening > 10 HTTP connections simultaneously, and keeping them all open for way too long.

https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/master/Shared/PlasmaDownloader/Packages/PackageDownloader.cs#L189

This logic is clearly wrong. Adding random delays won't fix it. You need to pipeline requests, you need to reuse connections, you need to close connections when they are no longer required.

If you are not ready to fix that, just remove the whole parallel thing... and pipeline everything.

Licho1:
[...]

PR #582, opened 2015-04-08
Package downloader anti spam tweak

xponen:
I remove the recent 3 sec anti-spam code from PackageDownloader.cs. This is because IMO its the caller that must avoid spam.

so I added 2 minute anti-spam code to PlasmaDownloader.cs (which is the caller of the Refresh). This period is same to one used by ZKL before ( as mentioned here: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/558 )

Additional change is: 
  - default engine to Spring 98.0.1-451
  - add 20sec cooldown timer to Reload button in Rapid tab to avoid spam

### DIFF
--- a/Shared/PlasmaDownloader/Packages/PackageDownloader.cs
+++ b/Shared/PlasmaDownloader/Packages/PackageDownloader.cs
@@ -30,11 +30,10 @@ namespace PlasmaDownloader.Packages
 
     public class PackageDownloader : IDisposable
     {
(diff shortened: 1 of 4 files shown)


Licho1:
It's  ok but why don't you fix it properly instead? Move it so that only server does the query and store rapid related data in DB instead.

xponen:
its awesome idea, but there's other stuff to do!

PR #583, opened 2015-04-08
Raises !y votes needed for !voteresign to 67%

Skasi:
Because there are way too many thrown games with the current 50% requirement.

Note that it's actually 67%+1 rounded down (provided thats how c# works), or:
2 people in 2v2
3 people in 3v3 and 4v4
4 people in 5v5
5 people in 6v6 and 7v7
etc.

Another option would be removing the vote's force-resign altogether (but keep the vote, it can help with communication), or only force-resigning !y voters.


Licho1:
I absolutely don't see why. Why it should be that 33% people who want to play block twice as many who don't.

Skasi:
They don't block anybody. People can still resign and quit the game.

xponen:
@Licho1  this commit don't violate that, if 67% want to play or resign, 33% must follow 67%

PR #588, opened 2015-04-13
Use module pictures from Github, not Google Code

sprunk:


Licho1:
Copy those to local server on deployment instead

xponen:
okay, Anarchid said same thing, but I didn't understand how it work. I don't know what Anarchid would do.

Anarchid:
I suspect xponen might lack admin privs on teamcity to do it.

sprunk:
module pictures are already on the local server
they are deployed using the avatar deployment, not game deployment though

Anarchid:
Excellent!

PR #589, opened 2015-04-13
Voting patch

sprunk:
* polls now start with !y vote from vote starter
* doing !command when not having rights for it but having rights for !votecommand will now run the vote variant, not the full variant (important because for example !move ignores room passwords, but !votemove doesnt).


xponen:
lol! exploit
merging asap

PR #598, opened 2015-04-19
Update spectator count when a spectator disconnects

db81:
This fixes negative player count in battle list.


Licho1:
good catch!

Issue #611, opened 2015-05-06
lobby: mods fail to download, outdated mods list


knorker:
lobby v1.3.2.62 which appears to be latest.

![bla](https://cloud.githubusercontent.com/assets/3044855/7492831/21d23a50-f3fc-11e4-8012-ff83afd8a07e.jpg)

When trying to host custom rooms, these mods failed to download:

Kernel Panic
Spring 1944
Gundam
(maybe more)

Some mods listed on https://springrts.com/wiki/Games are missing in the lobby's  selection menu:
Imperial Winter
On the Edge
Tech Annhilation
(maybe more)

The Cursed was not shown as "failed" but no progress happend and it reports 0B of 0B:

![bla2](https://cloud.githubusercontent.com/assets/3044855/7492966/4d63372c-f3fd-11e4-9aa5-6c3a7c434255.jpg)

downloading maps and zero-k worked.

knorker:
zkdota:stable failed too

xponen:
seem working for me, 
can you check/include ZKL diagnostic log for error?
![zkl 9_5_2015](https://cloud.githubusercontent.com/assets/2525365/7547206/2ce6720a-f61b-11e4-83b0-94ddf5345186.jpg)

xponen:
here's my rapid list:
is it missing the mod you mention?
![zkl_rapid_9_5_2015](https://cloud.githubusercontent.com/assets/2525365/7547233/cffae62e-f61b-11e4-83b0-d3820e1d26ca.jpg)

knorker:
The problem with "The Cursed" was maybe just random other problem.

But other mods:
Lobby tries to download s44:latest but in the rapid list there is only a s44:stable ?

![bla](https://cloud.githubusercontent.com/assets/3044855/7554275/5d988698-f721-11e4-9cd2-f67f364a3aad.png)

xponen:
okay I managed to trigger  s44:latest when creating new room for s44.
Its a UI issue, can be fixed.

Will add any missing games and check the tag (latest/stable/test) if it exist.

knorker:
thanks.
I could download KernelPanic now, but Spring1944 tries to download s44:stable which fails.
(even though it shows in that screenshot above, no idea)

PR #617, opened 2015-05-12
ZKL: update tag & add games to KnownGames 

xponen:
This fix https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/611

I included all games in https://springrts.com/wiki/Games (add OTE, Star Wars, Tech Anni) and update the rapid-tag of existing list (eg: "s44:latest" to "s44:stable")

These games appear in drop-down-list of ZKL's "Open New Battle" when "Official Only" isn't ticked.

Also
Hide ZK-Dota & Gundam because their rapid entry no longer exist

.


xponen:
seem like no controversy!

PR #618, opened 2015-05-13
Add startbox config to startscript as modoption

sprunk:
For lua reimplementation of startboxes - see https://github.com/ZeroK-RTS/Zero-K/issues/544

### DIFF
--- a/Shared/LobbyClient/Battle.cs
+++ b/Shared/LobbyClient/Battle.cs
@@ -295,7 +295,9 @@ namespace LobbyClient
                 userNum++;
             }
 
-            // ALLIANCES
+            // ALLIANCES AND START BOXES
+            var startboxes = new StringBuilder();
+            startboxes.AppendFormat("return { ");
             script.AppendLine();
             foreach (var allyNumber in
                 users.Where(x => !x.IsSpectator).Select(x => x.AllyNumber).Union(bots.Select(x => x.AllyNumber)).Union(_rectangles.Keys).Distinct())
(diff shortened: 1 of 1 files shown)


xponen:
"CultureInfo.InvariantCulture" is recommended to avoid unexpected numbering format.
eg: it will use exponent for fractional number or comma as decimal point based on server localisation. This make parsing from LUA impossible.

But
have you try using Spring callin? what's the benefit of duplicating same information that Spring already provided? both gadget & widget already have this information.

sprunk:
In the future Lua won't have this information through the engine startboxes as they will have to be removed (because they are uncontrollable).

Adding the CultureInfo stuff.

xponen:
@sprunk honestly its weird... if you're reimplementing startpos in LUA, wasn't it better to shuffle the startpos in a pre-game gadget and save result in GameRulesParam which all widget & gadget would use?

sprunk:
Engine boxes cannot be overridden through Lua so they need to go.

xponen:
No, you can override Engine boxes through Lua, and don't need this commit. 

Eg: Tell a gadget to put list of boxes in Global table, force others to use the table == same result as override.

sprunk:
Show me the Lua call-out that can control start-boxes

Anarchid:
> Method1: Tell gadget to put list of boxes in Global table, force others to use the table == same result as override. (benefit: localized to mod)

> Method2: Actual override Spring call: eg: https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Utilities/weapon_index_compat_91.lua (benefit: superb compatibility with past/future, no maintenance needed)

That is exactly similar to reimplementing them in Lua. You're arguing against your own position.

PR #621, opened 2015-05-13
Voteresign: don't count afkers for threshold

sprunk:
With an exception not to allow a single person to resign the whole team (takes care of most pathological cases).


xponen:
@sprunk is probably addressing the issue mentioned in https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/616

PR #624, opened 2015-05-16
ZKL: dont override clipboard by highlighting text

sprunk:
Undocumented, nonstandard behaviour that doesnt even ask the user before overriding the clipboard

### DIFF
--- a/ZeroKLobby/MicroLobby/TextWindow.cs
+++ b/ZeroKLobby/MicroLobby/TextWindow.cs
@@ -378,11 +378,6 @@ namespace ZeroKLobby.MicroLobby
                         buildString.Append(s);
                     }
 
-                    if (buildString.Length > 0) {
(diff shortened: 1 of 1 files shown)


Licho1:
Don't just remove until suitable alternative to copy pasting exists.
(Not having any copy paste is worse than having undocumented one)

Skasi:
Alternative already exists: Rightclick, open history, select text, ctrl+C.

Licho1:
Any ctrl+c action starts with left mouse select ,so in this respect it is relatively discoverable now with the tooltip alert.

sprunk:
It's mostly about overriding the clipboard even if the user does not explicitly want it

Licho1:
Then solve that part of the issue without removing feature. You could make it behave like standard text window with standard select (i know its not trivial, but its not especially hard either).

PR #629, opened 2015-05-20
Startbox config done properly

sprunk:


xponen:
Don't remove engine startbox, is used by other games

xponen:
cannot be merged, because break compatibility with other games, locking server to ZK only.

PR #631, opened 2015-05-22
xponen's startbox shuffler part 1

xponen:
added (to Springie) a cousin of !start: !startshuffled which invisibly shuffle player's startbox, then start the game.

Is very lightweight and very easy to understand. Completely independent of games, shouldn't cause much issue.

Related: https://github.com/ZeroK-RTS/Zero-K/issues/544

### DIFF
--- a/Shared/LobbyClient/Battle.cs
+++ b/Shared/LobbyClient/Battle.cs
@@ -295,6 +295,18 @@ namespace LobbyClient
                 userNum++;
             }
 
+            //Shuffle startbox
+            if (_modOptions.ContainsKey("shuffledbox"))
+            {
+                List<KeyValuePair<int,BattleRect>> shuffled = _rectangles.Shuffle();
+                int cnt = 0;
+                foreach (var keyValue in _rectangles)
+                {
+                    _rectangles[keyValue.Key] = shuffled[cnt].Value;
+                    cnt++;
+                }
+            }
+
             // ALLIANCES AND START BOXES
             // var startboxes = new StringBuilder();
             // startboxes.AppendFormat("return { ");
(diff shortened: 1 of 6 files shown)


sprunk:
It's "very lightweight" because it's only all-shuffle and has poor integration. You still need the normal shuffle and the whole game-side part.

It also fails horribly at ease of use  because everything (Springie, ZKL) assumes there is only !start so you'd also need to change all these to somehow decide which variant to use.

xponen:
!startshuffled+vote is 2x less step than doing !modoptions shuffle=xx+vote+!start+vote

Also, you can have a complex combo shuffling, (eg: !modoptions shuffle = startpos+ !startshuffled), at low coding cost.

sprunk:
You need to add the modoption to the game if the lobby is to set it.
If that exists then startshuffled is redundant (and inconsistent, you'd also need !forcestartshuffled etc.).
Your current implementation also fails because it enables shuffle regardless of whether the start actually passes.

xponen:
I think it's odd, but I have to thank you for finding these interesting flaws. I'll fix it and it will be the best shuffle ever!

GoogleFrog:
To me this looks like a crazy subversion of the modoption system. This is in ingame option so springie should not have change its behaviour depending on which setting is selected. The new type of start is inconsistent.

xponen:
changed how it work and fix issues raised by @sprunk 
- now a single option that work properly with vote
- is no longer haxing the modoption to signal shuffle, instead it is a Mode which Battleroom is currently in. If active: player are placed randomly to available boxes, if not, it follow player's allyNumber (normal)
- still add "shuffledbox=1" to modoption to signal Games (like ZK) but did nothing in Battleroom

xponen:
this is independent of games, it dont care if ZK reimplement startbox or not

PR #635, opened 2015-05-28
Placement matches dont show elo ingame

sprunk:


Licho1:
What is placement match?

sprunk:
Placement match is any match played when Elo is not yet fully known (= when Elo weight is not yet max).

Most games that use ranked matchmaking and put people into leagues (eg. SC2 or LoL) don't show any league before the rating is fully determined. This is good because:
* it helps other players determine someone who is new from someone who is just bad.
* doesn't show modified Elo, people get shown their "real" Elo as soon as their weight is maxed.
* it's consistent with how other games handle unknown rating so newbies know what's going on. It's more intuitive to tell them their rating is not yet known than to make their Elo grow rapidly which then suddenly stops.

This initial implementation prevents Springie from sending the Elo value to ingame, makes the playerlist show "???" instead of the number and use an "unknown" gray dude icon. In the future other ways of seeing Elo (dude icons in lobby, account page) would also hide Elo before it is fully known.

Licho1:
But it's not meant to show dudes or rank ingame.. just the new proposed colored combined level/elo bracket icon that will clearly signal newbies...

This work will be obsolete..

sprunk:
Ok I see.

Are the new combined elo+level icons available somewhere already? Replacing current dudes/ranks with them would not be hard and could be done even now.

Licho1:
It should not be needed,icon can be calculated form existing elo and rank

sprunk:
yeah i know, i mean are the new icon bitmaps available?

Licho1:
Luckywaldo promised to make them, there is a thread about it. He probably failed to deliver them but they can be generated using latest proposal, its simple thing.

sprunk:
ok will search for the thread

db81:
zero-k.info/Forum/Thread/3780?threadID=3780

sprunk:
Already found that but ty.

PR #638, opened 2015-05-29
Zero-K Multiplatform Installer

TurBoss:
works on windows xp, windows 7, and ubuntu
Needs testing

will add more distros


Anarchid:
This seems very interesting. 
More documentation about how it should be used?
What are the requirements, if any?

TurBoss:
this is an offline installer so the zklobby is included(an online installer requires the setup a repo witch is a xml plus some 7z containing the packages, can be any http location)
2 versions of it the .net 4.5.1 and th 4.0

the code is in /packages/info.zerok.installer/meta/

2 files are needed to edit the package.xml witch is a simple xml defining the package structure
and the installscript.qs witch does check for dependencies and runs the installer of them

There are 3 packages : 
The main package witch is selected by default the one for windows 7,
and the windows xp and ubuntu witch are selected depending on the OS found by the first package


For windows it checks for the same registers as xponen does to check if the .net versions are installed and runs the included web installer of each one depending on the windows version.

For ubuntu it checks for the binaries of all dependecies of zklobby and trays to install all from apt.
it also includes a sh witch runs "mono Zero-K" but i already have a .desktop file witch runs zk(it should be placed under ~/.local/share/applications/ (this works for ubuntu didn't know of other distros)

For add more distros i just know the simple name returned by "systemInfo.productType" and the add their respectibe package manager to the instructions followed by the installer to install the dependencies

so for finalize the work i need to know as many distros out there "systemInfo.productType"

thx

Licho1:
Thank you, few remarks:
 - created binary files should not be commited (add to gitignore instead)
 - will this support sort of packaging of selected folder? We will want to distribute game together with latest engine, maps etc, is this doable? Can it download that content optionally from website?

TurBoss:
Hi,

I'll remove de binaries from the repo

For additional stuff to be package just I need to create a new package for maps and another for the engine and select it for install depending on OS

the content of "packagename/data/ will be packed"

example:
we have this packages

info.zerok.installer
info.zerok.installerxp
info.zerok.installerlinux

then we just need to add

info.zerok.maps
info.zerok.enginewin
info.zerok.enginelinux

For online downloads a repository must be created by running the "buildrepo.bat" and place the output folder in a accessible http folder, and told the installer that the packages are now in the internet

Create a online installer
http://doc.qt.io/qtinstallerframework/ifw-online-installers.html
Promoting updates
http://doc.qt.io/qtinstallerframework/ifw-updates.html

TurBoss:
I close the PR
I will make a new one with the Licho suggestions

PR #644, opened 2015-05-30
Revert "Startbox patch"

xponen:
Reverts ZeroK-RTS/Zero-K-Infrastructure#641
this break Local game Skirmisher's compatibility with other games, and break server compatibility with other games.

The reason is simple, @Sprunk don't want engine startbox because it conflict with his LUA reimplementation of startbox.

This is bad, because startbox already work fine. If you want to shuffle it, just shuffle the content of startscript, not this.

### DIFF
--- a/Shared/LobbyClient/Battle.cs
+++ b/Shared/LobbyClient/Battle.cs
@@ -296,33 +296,34 @@ namespace LobbyClient
             }
 
             // ALLIANCES AND START BOXES
-            var startboxes = new StringBuilder();
-            startboxes.Append("return { ");
+            // var startboxes = new StringBuilder();
+            // startboxes.AppendFormat("return { ");
             script.AppendLine();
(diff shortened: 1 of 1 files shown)


xponen:
Both @GoogleFrog  & @sprunk say "its (startscript shuffling) already implemented".
Honestly, I have no idea what you guys were talking about.

If that's true, then why not just use it? why reimplement startbox to get a shuffle?
Try to explain what you mean by "already implemented"

sprunk:
The only remaining problem is that Springie crashes when attempting to generate startscript.
The logic is correct so it's just a matter of finding the crash.

xponen:
If you want Split startbox or Polygon startbox I can make it using Engine startbox by disallowing placement on dead region and adjust drawing

sprunk:
What you can do is help find the 1 crashing line

GoogleFrog:
If it crashes it should be reverted.

sprunk:
See #646

PR #645, opened 2015-05-31
Project files: add "AutoGenerateBindingRedirects" flag

xponen:
to project files. I think it is a good idea due to these VS warning (not error) that I got:
```
Found conflicts between different versions of the same dependent assembly. Please set the "AutoGenerateBindingRedirects" property to true in the project file. For more information, see http://go.microsoft.com/fwlink/?LinkId=294190.
```
After adding it, it fixed my "Newton.soft 4.5.0.0 not found" error on ZKL (net4.5.1): http://stackoverflow.com/questions/22507189/could-not-load-file-or-assembly-newtonsoft-json-version-4-5-0-0-culture-neutr

IMO its ZKLNet4.0's Newtonsoft fault but I didn't remember having to deal with this before , hopefully the flag is harmless and only make compiler more robust for all.

### DIFF
--- a/AutoRegistrator/AutoRegistrator.csproj
+++ b/AutoRegistrator/AutoRegistrator.csproj
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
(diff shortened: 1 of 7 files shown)


Licho1:
This is not the root cause, check packages one of the project is probalby using different newtonsoft version.
Make sure all project have the identical then there is no need for this,

xponen:
okay you are right, 
(from verbose build output) I found my VS is referring to a file from a duplicate ZK-infra. So weird, but fixed by deleting them

I guess AutoGenerateBindingRedirects wasn't entirely good, it might allow using old unintended stuff. I'll cancel this PR.

Ty

PR #646, opened 2015-05-31
Fix crash

sprunk:
Note: since it was not Springie then some other part of the infra still has 1-based boxes (perhaps the super obscure skirmish).


xponen:
can you add check for game named zero-k and allow legacy support for non-zk?

sprunk:
Eventually i guess

PR #647, opened 2015-05-31
remove wrong documentation

sprunk:
local testing _cannot_ and _should not_ be done on live db!

### DIFF
--- a/Springie/Springie/MainConfig.cs
+++ b/Springie/Springie/MainConfig.cs
@@ -45,13 +45,6 @@ namespace Springie
 		public int RepoMasterRefresh { get { return 120; } }
 		public string PackageMasterUrl { get { return "http://repos.springrts.com/"; } }
 
-        //Note for testing, 
-        //1) Set "ClusterNode" to "alpha"
-        //2) Make sure AutoUpdater() wasn't run (in Program.cs),
-        //3) Compile Springie & run Springie.exe
-        //4) Login to server (as player) using Lobby, 
-        //5) Join a room called "KingRaptor Secret Clubhouse",
-        //6) Use "magic" for password.
 	    public MainConfig()
 	    {
 	        ClusterNode = GlobalConst.SpringieNode;


xponen:
ZkServer should be coded to block rogue Springie using password instead of deleting comment to obscure debugging capability.

sprunk:
there is no secret password, you are just abusing your exception in the server firewall

xponen:
so it uses firewall.

so this is not needed.

db81:
Springe does not connect to the db and instead this method relies on http://zero-k.info/SpringieService.svc being completely open. That means that @xponen is not exploiting his exception and you can't cause a migration by running just springie (but what if you have an exception and accidentally run site together with it?).

But there are still good reasons not to use the obsolete method and test locally instead, namely you can't change the config without db access, it's inconsistent with how you test other components, it needs SpringieService to be accessible (it should be firewalled because it's very easy to abuse). The comment is an artifact of the time when local testing was impractical and now it's not useful and only serves to confuse.

Licho1:
yeah local testing can recreate whole env with site and springie correctly, it should indeed be removed

xponen:
just to add:
local testing is still not practical.
 
Some OS don't support website serving (eg: basic Windows), and to test require too much project to work together at same time without any fallback/limited setup (in case one of them fail) 
eg: I require website to test autohost?
 
Local test isn't as practical as said, the clubhouse is still needed.

Anarchid:
> Some OS don't support website serving (eg: basic Windows)

I'm no fan or defender of Microsoft, but this is FUD.

Licho1:
Its pure BS xponen. Vs - even the community one, comes with *two* different testing webservers. IIS express and WebDev, both are capable of running it all simply by selecting multiple start projects and hitting F5. No need to install anything or configure anything and no windows requirements (above VS)

xponen:
... everytime you guys says that, it make no sense at all, it never happen for me when I press F5.

Licho1:
Maybe got something like teamviewer and I can assist/check whats going on?

PR #650, opened 2015-06-08
Springie voting

sprunk:
Allow specs to vote on !votemove and !votesplitplayers
Allow specs to start and vote on !votemap


Licho1:
Why the should specs vote on map or split? They are not affected by it and should not vote.

sprunk:
Specs often are speccing because they dont like the map or want a smaller game and being able to vote on votemap/splitplayers makes them able to get what they want so they can play.

Also people always unspec -> vote -> spec back in lobby so preventing voting doesn't really work.

PR #655, opened 2015-06-14
Clean ZKL cache on first run

### LINKED ISSUE #654: Delete everyone path caches
I want ZKL to do a 1-time delete of everyones path caches. This is important and should be done ASAP.

KingRaptor:
Fixes #654

Note: This PR is going directly into stable! (to avoid having to cherry pick around welcome_screen merge)

### DIFF
--- a/ZeroKLobby/Program.cs
+++ b/ZeroKLobby/Program.cs
@@ -198,7 +198,41 @@ namespace ZeroKLobby
                 //};
 
                 SaveConfig();
-
+                if (Conf.CleanCache)
+                {
+                    try
+                    {
+                        var path = Program.SpringPaths.WritableDirectory;
+                        bool backupChatlogs = false;
+                        string sdPath = Utils.MakePath(path, "cache", "SD");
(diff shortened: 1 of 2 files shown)


ZeroKLobby/Program.cs
```diff
@@ -198,7 +198,41 @@ public static void Main(string[] args)
                 //};
 
                 SaveConfig();
-
+                if (Conf.CleanCache)
+                {
+                    try
+                    {
+                        var path = Program.SpringPaths.WritableDirectory;
```
Licho1: Imo you should only delete path cache.. other stores - like chat history, should be separate option

KingRaptor: It already preserves chat history (well I could give it an option to delete that as well, but that's not really within the scope of this change - I just need a temporary "delete paths" option).

Are there any other things I should try to keep as well? I noticed it also stores map/mod metadata and incomplete map (and maybe engine) downloads in the same folder.

Licho1: Ok I thought you would just delete cache/paths instead of this but.. why not..

PR #678, opened 2015-06-30
Add a diamond donator star

sprunk:


jhoblitt:
:+1: for same day service on a forum comment. :)  There's probably some optimal gamification value in the spacing between tiers.  I'd guess that increments of 500 or 1000 would be than a 4k jump as the later would require a fairly hefty donation by what I imagine is typical gamer standards.

KingRaptor:
Currently silver is 5x bronze, gold is only 2x silver, but diamond is 4x gold.
I'd probably move Gold to 2000 for more consistent spacing, so: 5x, 4x, 2.5x

jhoblitt:
My rough idea was to add several tiers at consistent spacing, with the spacing being close to the largest single donation a player would be likely to make. So something like:

1. 100
2. 500
3. 1000
4. 2000
5. 3000
6. 4000
7. 5000
8. ...

jhoblitt:
I wrote a [script](https://github.com/jhoblitt/zk-donations) to screen scrape http://zero-k.info/Contributions and compute some statistics.
```
mean: €30.68
median: €20.00
stddev: €45.06
min: €1.00
max: €406.00
percentile 10: €5.00
percentile 20: €10.00
percentile 30: €10.00
percentile 40: €15.00
percentile 50: €20.00
percentile 60: €20.00
percentile 70: €30.00
percentile 80: €46.77
percentile 90: €65.00
percentile 100: €406.00
```

Based on those numbers, I'd say that the tiers should have a stride of 500 points.

Skasi:
Or make it base 10!
1€
10€
100€
1000€

Licho1:
Yeah imo increase gold

jhoblitt:
17 of the 233 donations listed are >= 100€ or 7.2%.  I think that should be an upper band for the spacing between the tiers, which can always be rescaled at a later date.

Skasi:
Why reading through your forum post it sounds like we reeeally need a tier that has not been reached yet!
>  I'd like to suggest adding a couple of higher tier "Donator stars". I can't thank the Z-K devs enough, but I can try. :)

jhoblitt:
@Skasi That's exactly what I was/am suggesting.  I'm also suggesting that the spacing between the tiers shouldn't be exponential at this point in time. If you think there should be a "pie in the sky" 10,000€ reward tier, I'm all for it, but I also think that skipping from 100€ to 1000€ is probably poor [gamification](https://en.wikipedia.org/wiki/Gamification) of the reward structure.  The idea is to encourage people to donate and perhaps give a little more than they would otherwise just to hit the next tier.  A 900€ leap might actually be a disincentive to donate past the 100€ mark.

Kickstarter is a great example in that there are typically more than 3 tiers of "rewards".  How about we add a tier 4 at 1500 or 2000 kudos and wait to see what happens?  A 5th tier could always be added and/or the reward thresholds rescaled.

sprunk:
Rescaling upwards would require some retroactivity put in place or else some people would lose their icon.

How about:
* diamond at 250€
* silver moved 50€ -> 40€ for more even spacing

PR #686, opened 2015-07-10
Add engine and maps to multiplatform installer

TurBoss:
rewrite a lot of code

added maps and engine packages

next would be the online download but that requieres a working repo...


Licho1:
So how does it work? I can setup site to prepare "pack" in some place.
It should be two parts:
 - multiplatform part - maps, mod, various cache files
 - platform specific part - engine

Should I make site prepare some zip or something?

TurBoss:
to build a repo you need qt-ifw 2.0.1 once installed place maps ,engine and lobby in the correct folders , don't forget dotnet.
then run the buildrepo.bat and it should generate a folder caller repo can be changed editing the bat, the generated folder should contain some 7zips, host the repo folder is some http location then letme change the installer to make it an online installer

thx

Issue #695, opened 2015-07-13
Award images broken


Anarchid:
URL should be: http://zero-k.info/img/awards/trophy_air.png
URL is: http://zero-k.info/Battles/Detail/img/Awards/trophy_mex.png

-> url shouldn't be relative, url constructor should use whatever function required to get site root.

PR #696, opened 2015-07-13
Hardcode award URLs

sprunk:
URL construction is broken, see #695


Licho1:
Wrong, never hardcode stuff, site has to run under different domains, including locally.

PR #764, opened 2015-08-29
Special rights handling for selecting map from website

### LINKED ISSUE #756: (zklobby/host) Host says "no rights to change map" but then still does it
1) Go to a host, be alone in room
2) Click "maps" button in lobby
3) select map, click "Play on this map"
4) The host says:
* Elerium1 Sorry, you do not have rights to execute map
 * Elerium1 changing map to Small Supreme Islands V2

So it first says "No rights" but then still does change map.

KingRaptor:
Fixes #756

### DIFF
--- a/Springie/Springie/autohost/AutoHost.cs
+++ b/Springie/Springie/autohost/AutoHost.cs
@@ -234,7 +234,7 @@ namespace Springie.autohost
         }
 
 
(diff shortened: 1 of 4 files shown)


Licho1:
This is sort of odd now. Atm you can simply change it directly in the ZK server, without need for springie to cooperate

PR #768, opened 2015-08-30
Deploy cherries

Anarchid:
Cherry picks some fixes from previous attempted deployment while skimming over anything that looks dangerous and refactory. 

Goes directly into stable, to be merged during dead hours.


Licho1:
You should add steam handler crash fix, as it fixes typing zkl crash

PR #788, opened 2015-09-09
ZKL: write license files to datadir

KingRaptor:
Currently does GPL v3 and MIT License.


Licho1:
I hope steam wont freak out at the sight of this file in its folders..

PR #792, opened 2015-09-14
Limit ZKL error log size

KingRaptor:
Back up and clear error log if it exceeds 50 MB (overwrite any existing backup)


Licho1:
It's ok. For proper solution it could be swapped with trace and trace listener, because there are preexisting rolling backup trace listeners that do this.

Licho1:
Oh also technically speaking it can be executed from multiple threads at once - "delete" can fail because file is being written to atm if two exceptions happen at once on different threads.
I would recommend try catching delete and copy operation.

PR #793, opened 2015-09-20
next gen forum parser with wiki capabilities

Licho1:
Dont merge yet, some tags still missing


mojjj:
what does it do better? or was the old one broken? is the precompiled version stored in a seperate colum in the db now?

Licho1:
1) it's safe from various exploits and forum breakage
2) it supports nesting in places where it was previously impossible (nested quotes, nested spoilers)
3) it's faster

PR #800, opened 2015-09-23
Springie: Allow people to end their own votes

KingRaptor:


Springie/Springie/autohost/AutoHost.cs
```diff
@@ -700,7 +700,19 @@ public void Start()
             tas.LeaveBattle();
         }
 
-        public void StopVote() {
+        public void StopVote(TasSayEventArgs e = null) {
+            if (e != null)
+            {
+                string name = e.UserName;
+                if (name != null && activePoll != null && name != ((AbstractPoll)activePoll).Creator)
```
Licho1: Minor thing, if you add "Creator" to IVotable and make sure its always setup its slightly cleaner and no need to retypehere

PR #808, opened 2015-09-27
Add Serious AutohostMode

KingRaptor:
For things like Platinum and the new Unobtainium. Currently should be identical to teams host except blocks votes for special maps

[Rationale here](http://zero-k.info/Forum/Post/143065#143065)


Licho1:
Hmm same can probably be achived by tweaking balancer and nextmap picker. Those run on server and they know what host it is.. I mean probably not the right place to put this code ..

KingRaptor:
Votemap doesn't know whether a map is silly before it agrees to start the vote though.

PR #811, opened 2015-09-27
Serious host: block modoptions (zombies etc)

sprunk:


Licho1:
unable to merge, compilation fails, reopen when fixed

PR #812, opened 2015-09-27
Allow specs to vote on votesplitplayers and votemove

sprunk:
Also gives specs a response "only players can vote" if they try to vote on things they can't vote for (like kicks, map etc)


Licho1:
unable to merge, compilation fails, reopen when fixed

KingRaptor:
To fix the last commit add `using System.Linq;` to the top of the file

PR #829, opened 2015-10-11
Allow specs to vote on votesplitplayers and votemove

sprunk:
Also gives specs a response "only players can vote" if they try to vote on things they can't vote for (like kicks, map etc)

### DIFF
--- a/Springie/Springie/autohost/Polls/VoteMap.cs
+++ b/Springie/Springie/autohost/Polls/VoteMap.cs
@@ -112,5 +112,17 @@ namespace Springie.autohost.Polls
                 ah.ComMap(TasSayEventArgs.Default, new string[] { map });
             }
         }
+
+         protected override bool AllowVote(TasSayEventArgs e)
+        {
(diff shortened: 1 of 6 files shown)


Springie/Springie/autohost/Polls/VoteMap.cs
```diff
@@ -112,5 +112,17 @@ protected override void SuccessAction()
                 ah.ComMap(TasSayEventArgs.Default, new string[] { map });
             }
         }
+
+         protected override bool AllowVote(TasSayEventArgs e)
```
Licho1: Code quadriplication is ugly, perhaps add this to some abstract class from which these polls inherit?

sprunk:
Base class AllowVote returns true by default. If the no-spec-check was added to the base class then all the classes that allow specs would have to override instead.

Licho1:
You create class that inherits from base, and overrides this method. Then make those polls where vote is allowed inherit from this class instead of original base class.

sprunk:
Oh ok.

PR #859, opened 2015-10-19
CSS: .textbutton changes and small cleanups

Skasi:
.textbutton changes:

Added font-size:medium; to stop browsers from overwriting <input> fontsizes
Merged paddings to padding:0.4em 1em; to be the same as input.ui-button in jquery-ui.css
Added vertical-align:middle; to make sure <span> (default vertical-align:baseline) properly lines up with all the other buttons

Cleanups that don't affect styles, semantics unchanged:

In line 169 (now 197), shortened margin: 4px 1px 4px 1px; to margin: 4px 1px;
In line 191 (now 190), shortened padding: 15px 0 15px 0; to padding: 15px 0;
(see "The padding property can have from one to four values." on http://www.w3schools.com/css/css_padding.asp for a short explanation)
Also removed some empty lines, blank spaces at the end of lines and changed all blank space indents (usually four, sometimes two and five blank spaces) to use tab instead.

### DIFF
--- a/Zero-K.info/Styles/style.css
+++ b/Zero-K.info/Styles/style.css
@@ -2,8 +2,8 @@
 
 html
 {
-    /*background: #000 url("/img/bg_bluehex.gif") fixed;*/
-    background: #000;
+	/*background: #000 url("/img/bg_bluehex.gif") fixed;*/
+	background: #000;
 	color: #ede;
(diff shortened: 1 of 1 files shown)


Skasi:
I should note that this change makes some buttons with borders slightly larger.

PR #862, opened 2015-10-20
added adminlink to usersearch

Jamanno:
slap Chesti if it breaks usersearch


Licho1:
non admins should not see that

PR #863, opened 2015-10-20
added adminlink to usersearch

Jamanno:
without that link you need to click 2 times until you can see admin relevant data.


Zero-K.info/Views/Users/UsersIndex.cshtml
```diff
@@ -9,6 +9,10 @@
     grid.AddCol("ID", x => x.AccountID).SetSort(x => x.AccountID);
     grid.AddCol("<span nicetitle='Is admin?'>A</span>", x => x.IsZeroKAdmin ? "<img src='/img/police.png'/>" : "").SetSort(x => x.IsZeroKAdmin);
     grid.AddCol("Name", x => Html.PrintAccount(x)).SetSort(x => x.Name);
+    @if (Global.IsZeroKAdmin)
```
Licho1: @ does not belong there.. did you test it?
You use @ if you switch from html to c# or back

PR #882, opened 2015-12-03
Replace "Social Media" logos with simpler/non-404 resources

Skasi:
The current version looks pretty weird and is hard to read on the near-black background. This change should ensure (new) visitors notice the Github symbol. Here's what the new image looks like: http://i.imgur.com/c1J59Dy.png

I took the new logo "GitHub-Mark-Light-32px.png" straight from https://github.com/logos


Licho1:
I don't see any loo on that URL it appears wrong

Skasi:
Uh no. Nothing's wrong, everything just like it's supposed to be.
Which url do you have a problem with?

Skasi:
Added a fix for the steam logo (returned 404).

PR #924, opened 2016-01-08
Server support for dynamic comms

KingRaptor:


Licho1:
You modified data class to maked sure open package manager in VS and type
add-migration somename
it should generate migration files for db

PR #988, opened 2016-03-30
Only allow custom modoptions on private hosts

sprunk:
This is done by removing !votesetoptions
Bosses of private rooms can still use !setoptions for a custom game

### DIFF
--- a/Springie/Springie/autohost/CommandList.cs
+++ b/Springie/Springie/autohost/CommandList.cs
@@ -171,11 +171,7 @@ namespace Springie.autohost
 
             AddMissing(new CommandConfig("setoptions", 3, "<name>=<value>[,<name>=<value>] - applies mod/map options", 0));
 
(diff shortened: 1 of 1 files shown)


Licho1:
I'm not sure about this. What do other people think?>

knorker:
What use is there for modoptions?
The potentially interesting gamemods were never used because lobby made it too difficult to set up and everything was half-broken, until it got removed.

The only modoption I ever see is _metalmult_  which is is used to:
1) to create silly nonsense spam games with +9999 income
2) make tiny maps with few mexes 'more playable' for clusterfuck sized teams.
Both is silly and offputting.

The metalmul modoption is not nessecary to fix bad maps because since years that can be done with lua-metal config file.

CrazyEddieTK:
This should be adopted.

Skasi:
For clarity the Options button in public rooms should be hidden, grayed out or the Options menu - when accessed - should inform players about this once it goes live.

Licho1:
Its being done in new springie-less sytem

Issue #921, opened 2016-01-01
autohost settings produce unattractive matches


knorker:
This is about the official, public autohosts.

1) map selection
There exists system of "featured" maps and definitions if a map is suitable for 1v1, teams or FFA. Why is it not used? 
-Tiny maps should not be used in large-teams room because result is choatic Commander-deathmatch. 
-Joke-maps like: Trololo, Duke-Nukem or maps that by usual criteria would never get the "featured" (or even "supported") tag like superlong variants of IcyRun. Those should not be in any room.

2) Team sizes are too large.

3) "Funny" settings.
For example superhigh income-multiplicator so that within few seconds the income is +50metal.
It breaks every aspect of balance. This has no place in official room.

4) Broken maps or 'testing' maps.
Official rooms are a bad place to test maps.

Skasi:
I think what you really want is matchmaking.

knorker:
No.
This is about how autohosts should be configured, so players can meet there to play the game in a normal way. (If you want to discuss what "normal" means, then there are already plenty of threads on the forum to do excactly that.)

Skasi:
Forum posts and polls have shown that players do not like player-limits. They also prefer playing a 10v10 on icy run over a 20x20 map that reduces FPS to a single digit.

Joke maps can **definitely** be supported. They often don't break any game mechanics, don't crash the game and don't cause any other technical problems. There's a specific "Silly" tag for joke maps, so neither "featured" nor "supported" need to be reserved for non-joke maps only.

>  It breaks every aspect of balance. This has no place in official room.

 &
> This is about how autohosts should be configured, so players can meet there to play the game in a normal way.

I don't think this is how rooms titled "All welcome!" are supposed to work. I am pretty sure I understand what you want: You want some sort of "competitive"/"serious" room, but just because a room has a star showing it's "official", doesn't mean it's supposed to be either of those. Official means "this room is not hosted by a private individual, it is public, vote-based and fully subject to moderation (ie. CoC applies)".

While there could indeed be a room titled "Competitive (team) games" or similar I think going straight towards matchmaking would remove issues like !vote spam, spec cheating and similar that all come with the current room-system and would affect competitive/serious rooms more negatively than others.

knorker:
I do not think that this is very "competetive" or "serious" or "hardcore" or whatever it gets called sometimes.
It is simply the normal way. 
metalmulti=1.0 is default, so that is normal.
[...]

PR #989, opened 2016-03-30
Map pool stuff

sprunk:
* 1v1 host only allows maps tagged as 1v1, analogical for teams and FFA
 * Unfeatured and silly maps are only allowed on private hosts

### DIFF
--- a/Springie/Springie/autohost/AutoHost_commands.cs
+++ b/Springie/Springie/autohost/AutoHost_commands.cs
@@ -380,7 +380,7 @@ namespace Springie.autohost
             {
                 SayBattle("Changing map to " + vals[0]);
                 var mapi = cache.GetResourceDataByInternalName(vals[0]);
-                if (mapi != null && mapi.MapIsSupported == true)
+                if (mapi != null)
                 {
                     tas.ChangeMap(mapi.InternalName);
                 }
--- a/Springie/Springie/autohost/AutoHost_commands.cs
+++ b/Springie/Springie/autohost/AutoHost_commands.cs
@@ -378,9 +378,9 @@ namespace Springie.autohost
             int[] indexes;
             if (FilterMaps(words, out vals, out indexes) > 0)
             {
-                SayBattle("changing map to " + vals[0]);
+                SayBattle("Changing map to " + vals[0]);
                 var mapi = cache.GetResourceDataByInternalName(vals[0]);
-                if (mapi != null)
+                if (mapi != null && mapi.MapIsSupported == true)
                 {
                     tas.ChangeMap(mapi.InternalName);
                 }
--- a/Springie/Springie/autohost/Polls/VoteMap.cs
+++ b/Springie/Springie/autohost/Polls/VoteMap.cs
@@ -32,20 +32,40 @@ namespace Springie.autohost.Polls
                     ah.FilterMaps(words, out vals, out indexes);
                     if (vals.Length > 0)
                     {
-                        bool serious = ah.config.Mode == AutohostMode.Serious;
-                        var db = serious ? new ZkDataContext() : null;
+                        var db = new ZkDataContext();
 
                         foreach (string possibleMap in vals)
                         {
-                            if (serious)
+                            try
                             {
-                                try
-                                {
-                                    var mapEntry = db.Resources.FirstOrDefault(x => x.InternalName == possibleMap);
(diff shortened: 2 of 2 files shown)


Licho1:
disagree, its up to people to decide. Its not in default mapcycle, if people do want those maps, they should be allowed to pick them

Skasi:
> Non-supported are not allowed even in private

Definitely against this change. Private games shouldn't have any restrictions.

sprunk:
Removed the "no unsupported maps" (the solution to swastika map spam is not to show private rooms by default in lobby)

db81:
Another solutions is to cycle the map to the next featured when the room becomes empty. That fixes trololo/duck/cow spam as well.

sprunk:
Then you can just keep 1 smurf inside.

knorker:
> Another solutions is to cycle the map to the next featured when the room becomes empty. That fixes trololo/duck/cow spam as well.

No, that is another topic.
The problem is that when you join a room you can never be sure if there will not suddendly be a mapchange to trololo/duck/cow. It is annoying to hope for an RTS match, invest waiting time, then tada! There comes the singing fuckface map.
This PR would be a first step towards at improving https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/921 but sadly some admins block any attempts at fixing the degraded nonsense culture that has developed.

CarabusX:
You are trying to severely limit the ability to test new maps for everyone, including mappers, just because some troll made swastika rooms a few times. This is would mean the troll wins, just like terrorists win IRL when politicians start limiting the freedom of everyone.

sprunk:
>severely limit the ability to test new maps

tagging a map as supported takes literally 10 seconds

CarabusX:
> tagging a map as supported takes literally 10 seconds

But would have to be done to 10000+ maps most of which are valid and playable, just because there are 2 or 3 troll ones. Some others are bugged too, but your point won't apply either, because testing these is definitely not 10 seconds,

Skasi:
> tagging a map as supported takes literally 10 seconds

Nope, it's been a month. http://zero-k.info/Maps/Detail/51571

Relying on inconsistent people to be available and not too lazy to do something is a bad idea.

knorker:
> >tagging a map as supported takes literally 10 seconds
> Nope, it's been a month. http://zero-k.info/Maps/Detail/51571 

No match played on the map, doubtful if someone even tested it offline, no nothing, on what facts should the map be tagged supported? Because you posted:  "Very boxxy." ?

knorker:
also github is as bad trollground as forum now, grats.

sprunk:
>But would have to be done to 10000+ maps most of which are valid and playable

Wouldn't have to. They are so "valid and playable" that 9000 of them have never seen a single game. They are utter shit and nobody actually wants to play them.

Anyway it would be trivial to just mass-support all existing maps which had >0 games on them and pick out the few evil ones. If someone happens to find a valid old map with 0 games it's effectively the same as creating a new one - test it offline first, getting it tagged is the least concern (Skasi just sucks at talking if he couldn't get a map tagged for a month).

Skasi:
> No match played on the map, doubtful if someone even tested it offline

Remember test games that are !exited do not get recorded.

> on what facts should the map be tagged supported?

On facts gathered through proactive science by people responsible for tagging. Or perhaps whatever sprung was thinking.

sprunk:
If there is an existing PvP match on the map, just point to it.

If not, download it and play locally. If it works then contact an admin through PM, report, or forumpost under map (tip: on forum prefer things like "this map works and is playable" instead of "very boxxy").

knorker:
@GoogleFrog

Skasi:
I think before this PR gets accepted someone needs to make sure that things like FFA maps which work fine for team games will be allowed in team rooms. http://zero-k.info/Maps/Detail/7133 comes to mind.

CrazyEddieTK:
Now that the PR has been fixed so that it only applies to public rooms and not private rooms, it should be adopted.

Although isn't it already partially adopted? Isn't it the case that right now non-supported maps can't be chosen in public hosts? If so, then the other part that needs to be adopted is to block special maps from being chosen in public hosts.

sprunk:
No, you can choose any map everywhere currently.

Licho1:
Its now being done by specifying host type when hosting, custom = no restriction, otherwise only proper maps + no modoptions.

PR #1001, opened 2016-04-15
fix #967

### LINKED ISSUE #967: (lobby) better handling of non-downloadable maps
There are hundreds of non-downloadable maps.

1) When using partial mapnames in !map commands then only downloadable maps should be picked.
For example:
>  !map desertneed

This should NOT pick the undownloadable version ( http://zero-k.info/Maps/Detail/20875 ) but instead a downloadable version ( http://zero-k.info/Maps/Detail/51570 )

2) The !maplink command shows that no download-mirrors exist.
Yet the lobby tries forever to download the map, until manually canceled.

maplink does not exist

sprunk:
only return undownloadable maps on exact match

### DIFF
--- a/Zero-K.info/ContentService.svc.cs
+++ b/Zero-K.info/ContentService.svc.cs
@@ -38,6 +38,7 @@ namespace ZeroKWeb
                     ret = ret.Where(x => SqlFunctions.PatIndex("%" + w1 + "%", x.InternalName) > 0);
                 }
             }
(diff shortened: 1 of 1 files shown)


Licho1:
This makes no sense, there can be some files which have 0 links and some that have links. This blocks if tehre is single file with no links, reverting

sprunk:
change to Any insteal of All?

PR #1006, opened 2016-04-15
1v1 elo now uses weight

sprunk:


Licho1:
???? it makes no sense mathematically! It's zero sum!

sprunk:
how about this: #1008

PR #1008, opened 2016-04-15
1v1 weight done the other way

sprunk:


Licho1:
You cannot use effective elo it would lead to non-zero sum because of malus...

sprunk:
That would be only for the win chance calculation (like it already is for teams, see https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/master/ZkData/Ef/SpringBattle.cs#L128 ).

Licho1:
ok

Licho1:
I mean we can keep it this way i suppose.

sprunk:
"This way"? This here is a PR not a commit.

PR #1050, opened 2016-05-23
Change GetEloTop10() to conform with the ladder

ashdnazg:
same query as in:
https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/master/Zero-K.info/Controllers/LaddersController.cs

should affect the little trophies in the lobby player lists


ashdnazg:
hrumph

PR #1062, opened 2016-06-02
Deploy to alleviate Springie not noticing ZK releases

Anarchid:


db81:
That only solves part of the problem, they still need to be restarted before they update. The fix is that you don't have to manually !rehost them.

Anarchid:
Yes. Hence 'alleviate' and not 'solve'.

Some bonus tier madness: playing a manually !spawned game with new stable registers it into Resources. But springies still do not update on !rehost, or even on !updaterapidmod !rehost.

PR #1071, opened 2016-06-23
Highlight posts that have been linked to.

Skasi:
When using the "Link to this post" to create an url that contains an #anchor the post the anchor points to will be highlighted with a slightly dark teal background.

Example url: http://zero-k.info/Forum/Thread/14261?postID=159558#159558

Example screenshot


Skasi:
To clarify, this is to make it clear which post someone is referring to. That should be particularly useful when a post is close to the end of a thread and does not end up at the very top of the screen when jumped to using an anchor.

PR #1087, opened 2016-07-07
adjust icons

Anarchid:
replaces ugly multiplayer icon
add alternative web icon (not applied; discuss)


KingRaptor:
I like the web icon.

The new MP icon is good but I find myself oddly attached to the conjoined twins >__<

db81:
The twins read more like "play with friends" than MP, so if you're attached we could perhaps use them somewhere in that vein.

Licho1:
Yeah i would like to use MP for overlay, and keep current for button, need to cherry pick this later.

Licho1:
ZKL now uses these icons (merged manually to avoid overwrite of MP)

PR #1133, opened 2016-08-22
Add alt text for search buttons

Skasi:
Inputs of type image require an attribute for alternative text. This makes buttons usable when images are not loaded (eg. when disabling them to store bandwidth on mobile, or when the file gets lost for whatever reason).


Skasi:
Feel free to [squash and merge commits](https://help.github.com/articles/about-pull-request-merge-squashing/)

PR #1138, opened 2016-08-29
Fix userspoofing in admin page

ashdnazg:


Zero-K.info/Controllers/UsersController.cs
```diff
@@ -87,7 +87,7 @@ public ActionResult ChangeElo(int accountID, int eloweight, int eloweight1v1)
         {
             var db = new ZkDataContext();
             Account acc = db.Accounts.Single(x => x.AccountID == accountID);
-            Account adminAcc = Global.Account
+            Account adminAcc = Global.Account;
```
ashdnazg: ah, thanks :)

Licho1:
recreated as fixed pr

Issue #1200, opened 2016-09-11
Persist MM Rooms


DeinFreund:
Some 30 seconds after each match for gg would be nice. Also maybe as a chance to use a future rematch command.

This would also give some time to look at the post-game stats.

GoogleFrog:
Easy access to asking for a rematch creates social pressure that people may not want. How do both players feel if one asks for a rematch and the other declines? Perhaps they feel too awkward to requeue. I think easy rematch runs counter to the aim of having many people play a variety of opponents.

I don't know what you mean by "MM Rooms" because post game stats are not viewed in what I would call a room. Also, there is plenty of time to view the stats even if the server stops. However, I do support running the server for a short while after the game ends for things like 'gg' and perhaps advice.

DeinFreund:
The problem is just not being able to use the chat once the server shuts down. That's where I'd like the little extra time.

GoogleFrog:
In that case I think it would be best to have the springie host remain online for a minute after the battle ends.

Licho1:
I think its a bad idea, people might do stuff meanwhile, like start it..

GoogleFrog:
My plan for the UI is to hide the existence of the MM battle rooms. I hope these battles have a tag along the lines of "matchMakerBattle = true". People will be able to click on these battles in the "Watch" submenu of the MP menu but they will be unable to join them.

PR #1202, opened 2016-09-12
Spring lives for a minute after game-over

sprunk:
See https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/1200#issuecomment-246244631


Licho1:
why would you want this?

Licho1:
I added internal wait to serverbattle instead of keeping spring alive, as was recommended by google (keep lobby alive not spring)

sprunk:
Oh, I misunderstood "host remain online".

PR #1240, opened 2016-09-17
new ladders -> channelmanager

DeinFreund:
shouldn't this be the same for both?


Licho1:
oh it would have worked, i only now found your pr. More universal solution already in place (generates and caches it just in one place instead of copying code)

DeinFreund:
That's the proper solution anyway.

PR #1242, opened 2016-09-17
removed yellow/orange intermediary shade

DeinFreund:
there has been a lot of criticism about how hard it is to differentiate this shade. 

the brackets have been shifted down to <1300, 1300-1500, ..., 2100-2300, 2300+
if 2400+ is needed again, we could just increase the size of the brackets.


Licho1:
I don't think its needed, i like orange. Voting for keping it.

DeinFreund:
Orange is still there, it's now Red->Orange->Yellow instead of Red->Orange->Dark Yellow->Bright Yellow

DeinFreund:
This also makes the site use unshaded icons, any critic regarding that point?

GoogleFrog:
The site should use the shading style that it currently uses. I think the bracket changes are reasonable given the restriction. I would prefer it if another colour could be added to maintain the current brackets.

Licho1:
I think that there is a room on the other side of spectra - adding some brownish shade between gray and red. 
It's not a black body radiation but should be intuitive where it belongs.

Anarchid:
There's a lot of room on the higher end of the spectrum too: magenta, violet, ultraviolet.

DeinFreund:
Brown was an idea before but it doesn't work well because it's too dark. Purple is already the highest shade, but if you think you can make another easy to understand shade around that hue go ahead.

sprunk:
Looks like it was decided not to do this and there is also a conflict.

PR #1250, opened 2016-09-18
unshaded icons for zk.info

DeinFreund:
Alternative to #1242


DeinFreund:
Seems some want it the other way around.. I guess we'll just leave it be.

PR #1360, opened 2016-12-20
Show "correct" (competitive) elo changes on battle stats page

DeinFreund:
It was showing the casual elo changes on MM battle pages. If both elo changes are supposed to be displayed (for admins), player.EloChange needs to be replaced by a Dictionary with keys  for each elo type.

### DIFF
--- a/ZkData/Ef/SpringBattle.cs
+++ b/ZkData/Ef/SpringBattle.cs
@@ -84,8 +84,8 @@ namespace ZkData
             }
             else
             {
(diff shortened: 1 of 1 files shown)


Licho1:
Your change does nothing. You just changed order of lines but it does nothing.

DeinFreund:
So where is player.EloChange set? To me it looked like the second call to elo calculation overwrites what is set in the first one.

Maybe it's easier if you fix it.

Licho1:
Damn, you are right... such a stupid bug/logic.. this thing should be a parameter or set elsewhere.. oh well thanks for finding!

PR #1363, opened 2016-12-22
Site: GetUserIP() returns an IPv4 if present

KingRaptor:
This fixes a potential failure with site ban by IP, where it currently gets the IPv6 and tries to look for matching IPv4 bans.

(Alternatively lobby clients could be allowed to report IPv6 instead)

### DIFF
--- a/Zero-K.info/Global.asax.cs
+++ b/Zero-K.info/Global.asax.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Net;
 using System.Web;
 using System.Web.Mvc;
 using System.Web.Optimization;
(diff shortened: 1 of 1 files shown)


Licho1:
This returns IP address of the server instead of the IP address of the person who made the request. You are resolving empty string.

KingRaptor:
Is there a way to access `MvcApplication` from `ForumController`? So forum ban check (e.g. [here](https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/52cbec2034bc459f7fdedad40aa7374fe4fa3466/Zero-K.info/Controllers/ForumController.cs#L247)) can get the correct IP as well.

Licho1:
ForumController has "Request" object in it, so it can extract data from it. This function could be written in some other place and take Request as an input parameter instead.

Licho1:
I'm still not sure about this thing. Running DNS resolve will slow down the site a lot. 
Option could be to simply disable IPv6 on the server.

Licho1:
I think if it only checked ipv6 and not all it should be fine..
use something like here?
http://stackoverflow.com/questions/1932843/iis-request-userhostaddress-returning-ipv6-1-even-when-ipv6-disabled

Issue #1373, opened 2017-01-03
Free up user names


GoogleFrog:
Old usernames can be freed up by renaming users from Name to OldUser<nonsense>Name. Here are some decent criteria for the process:
 * Most recent login or forum post occurred in 2014 or earlier.
 * Less than 10 games played.

Licho1:
this is going to free 33366 account names, which is pretty good.

GoogleFrog:
How many account names are there in total?

Licho1:
About 70 000

KingRaptor:
While we're at it I'd like to rename all accounts with case-sensitivity conflicts as DeinFreund suggested (non-oldest accounts get _1, _2, _3, etc. appended to end of name) and then make name collision check on registration case insensitive.

Licho1:
Renaming accounts is more disruptive it will affect existing players... what is the exact rule?
Regarding old accounts i will just rename "nickname" to "nickname#old#accountid"

db81:
We can use latest logged in instead of oldest to resolve case-sensitivity.

DeinFreund:
We could also just disallow "duplicate"(matching an existing name ignoring case) registrations from now on. As long as duplicate names are only a problem to humans reading them this should be the easiest solution.

Licho1:
Yeah thats probably the best thing to do .. case conflicts are only issue for humans.

PR #1378, opened 2017-01-13
Case insensitive search for existing usernames during account registration

KingRaptor:
See https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/1373#issuecomment-270851110


Licho1:
Is this really working? It will probably result in index scan (since index is case insesitive) but query is rare enough to not matter.

KingRaptor:
Seems to not actually be case insensitive, I'll have to investigate further and fix.

PR #1484, opened 2017-02-20
Add java checking methods[wip]

DeinFreund:
This is just the methods without using them as I don't know the protocol between Chobby<->Wrapper

### DIFF
--- a/ChobbyLauncher/Chobbyla.cs
+++ b/ChobbyLauncher/Chobbyla.cs
@@ -161,6 +161,11 @@ namespace ChobbyLauncher
 
         private bool CheckForJava()
         {
+            int p = (int)Environment.OSVersion.Platform;
+            bool isLinux = (p == 4) || (p == 6) || (p == 128);
+
+            if (isLinux) return false;
+
             string[] envVars = { "JAVA_HOME", "JDK_HOME", "JRE_HOME" };
 
             foreach (string envVar in envVars) { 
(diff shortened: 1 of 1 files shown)


ChobbyLauncher/Chobbyla.cs
```diff
@@ -161,6 +161,11 @@ private dynamic ExtractEngineFromLua(PackageDownloader.Version ver)
 
         private bool CheckForJava()
         {
+            int p = (int)Environment.OSVersion.Platform;
+            bool isLinux = (p == 4) || (p == 6) || (p == 128);
```
sprunk: magic -> named consts?

DeinFreund:
This only works for windows where spring is always 32 bit for now. :/

I'm not sure if it's even worth trying on Linux because Spring can be really picky with the java.

ChobbyLauncher/Chobbyla.cs
DeinFreund: http://www.mono-project.com/docs/faq/technical/

That's the recommended way of checking.. But I can name them if you want.

sprunk: I don't want.

DeinFreund: Ok, because it's not going to make this any less magical to me.

Licho1:
will go to engine

DeinFreund:
Doesn't seem

PR #1485, opened 2017-02-20
Report dev status + donation amount to ingame

DeinFreund:
Not sure if I should just change DB like this, but I think we need a new bit if we want to reward devs.

### DIFF
--- a/ZkData/Ef/Account.cs
+++ b/ZkData/Ef/Account.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.ComponentModel.DataAnnotations.Schema;
@@ -128,6 +128,7 @@ namespace ZkData
         public double PwWarpUsed { get; set; }
         public double PwAttackPoints { get; set; }
         public bool HasVpnException { get; set; }
+        public bool HasDevStatus { get; set; }
         public int Kudos { get; set; }
         public int ForumTotalUpvotes { get; set; }
         public int ForumTotalDownvotes { get; set; }
--- a/ZkLobbyServer/SpringieInterface/StartSetup.cs
+++ b/ZkLobbyServer/SpringieInterface/StartSetup.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Data.Entity;
 using System.Diagnostics;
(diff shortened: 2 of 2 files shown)


sprunk:
probably needs migration for existing DB?

DeinFreund:
Indaled

ZkLobbyServer/SpringieInterface/StartSetup.cs
```diff
@@ -92,6 +92,13 @@ public static LobbyHostingContext GetDedicatedServerStartSetup(LobbyHostingConte
                         userParams["elo"] = Math.Round(elo).ToString(); 
                         userParams["skill_order"] = ((context.IsMatchMakerGame ? user.CompetitiveRank : user.CasualRank) ?? int.MaxValue).ToString(); // send order of skills (For lists). Note this should be improved by sendng normalized list instead of ranks
 
+
+                        var contribs = user.ContributionsByAccountID.ToList();
+                        var total = 0;
+                        foreach (var contrib in contribs) total += contrib.KudosValue;
+                        userParams["donations"] = total.ToString();
```
sprunk: user.KudosGained ?

DeinFreund: Is this the same? Then sure. This was taken from user page code.

Licho1: Use value from Account directly

ZkData/Ef/Account.cs
```diff
@@ -128,6 +128,7 @@ public Account()
         public double PwWarpUsed { get; set; }
         public double PwAttackPoints { get; set; }
         public bool HasVpnException { get; set; }
+        public bool HasDevStatus { get; set; }
```
Licho1: if you do changes in DB you need to do following: 
open nuget console, set project to "ZkData" and write:
add-migration AddedDevStatus
This will prepare classes to evolve DB.

Also dev status isnt 1/0 it should be enum (none, contributor, developer, coredev) or something like that maybe?

DeinFreund: i could just make it an int so we're open.

if you want you and all the long time peoplers can have extra core status, but i'd not be too greedy. we want to encourage everyone to try to contribute after all and it doesn't cost us anything to give out.  also the more categories the harder it'll be to draw the line and to keep it updated. discuss on http://zero-k.info/Forum/Thread/23637

sprunk: To consider: differentiate game contributor (things to ZK repo itself) and external content contributor (making maps / AI / missions etc)?

Licho1: Enum is mapped to int in a database. 
And yes idea was:
contributor (does stuff, maps, random PR)
developer (has push access)
coredev

DeinFreund:
Licho maeks it

ZkData/Ef/Account.cs
ashdnazg: what about engine devs :P

DeinFreund: u get a laureated lobster, happy?

ashdnazg: I want to see!!!

ashdnazg: and yes :)

PR #1537, opened 2017-03-02
unreadable as it gets

DeinFreund:


Licho1:
this is too much, if you want to go one letter, you might as wel send | separted liste instead of json structure

DeinFreund:
That'd work as well

PR #1555, opened 2017-03-06
[WIP] Whole History Rating

DeinFreund:


DeinFreund:
I'll need some help in getting some local replays to test this on.

Licho1:
It should integrate with existing systems.. Look how generic calculate elo is, it's a generic function that is reused for both MM and non-mm elo. Can something else be provided? This system as is is basically separate thingie floating on the side.

DeinFreund:
I will provide an interface similar to Elo calculation. All you'll have to do is a call to calculate WHR after each battle. I'll add such a method next to calculateallelo. Same thing goes for reading values.

I'm keeping the actual implementation out of the existing namespace as it would just crowd it up with classes and methods that are irrelevant to the rest of zk.

DeinFreund:
This was just the raw port from my previous implementation in Java.

DeinFreund:
I've made a general interface for rating systems and generalized categories(Casual, MM, PW). This interface could also be used for Elo or other future systems and allows direct comparisons to be made. (It's similar to what I used locally to compare all those different rating system suggestions)

I'll need some more direct feedback when it comes to development principles for infra. CalculateEloGeneric and other Elo functions are all fixed to Elo, it's not very generic at all to redo these method for every rating system.

PR #1565, opened 2017-03-08
Low time resolution to see if that's causing oom

DeinFreund:


DeinFreund:
For reference, 1 day resolution went oom after 92k processed battles.

Licho1:
i dont think this helps, probably its just too many battles kept in memory at the same time.

PR #1575, opened 2017-03-08
Elo replacement proof of concept (disabled on live)

DeinFreund:


Licho1:
This wont work, compiled expression is basically to create shortcut for computet properties, but properties computed on sql server. 
This code could only work in memory..

PR #1584, opened 2017-03-10
Safety for comm modules

sprunk:


Licho1:
It would have crashed otherwise (single crashes if there is no entry or more than one), but this is more explicit)

PR #1616, opened 2017-03-14
fix possible null pointer

DeinFreund:


ZkData/Ef/WHR/PlayerRating.cs
```diff
@@ -17,7 +17,7 @@ public class PlayerRating
         public readonly int Rank;
         public float Uncertainty { get
             {
-                return UncertaintyFunc.Invoke();
+                return UncertaintyFunc != null ? UncertaintyFunc.Invoke() : float.PositiveInfinity;
```
sprunk: `return UncertaintyFunc?.Invoke() ?? float.PositiveInfinity;`?

DeinFreund: make pr

PR #1623, opened 2017-03-14
Safety for planet owner assignment

sprunk:
No idea if this changes logic but modifying the value after queuing it for insert looks suspicious.


Licho1:
this makes no difference at all...

PR #1630, opened 2017-03-14
fix destroy null pointer?

DeinFreund:


Licho1:
this would fix it but crash on next linm wher eit access .Account whihch isnt preloaeded either, i will fix it

DeinFreund:
Next line has a null check though

Licho1:
So it would never give any metal to any player despite claiming to do so ..

DeinFreund:
Great isn't it?

PR #1631, opened 2017-03-14
dont allow attacking a planet that's being fought for

DeinFreund:


Licho1:
1) not a correct place/only place, proper place is CanMatchMakerAttack 
2) do we want this? This would lmit number of people who can play

DeinFreund:
There have been problems with fighting for a planet that has already been captured. I.e. invading a planet that is already yours.

DeinFreund:
CanMatchMakerPlay has no access to Global

PR #1647, opened 2017-03-16
Clan leader can delete posts from his clan thread

sprunk:


Zero-K.info/Controllers/ForumController.cs
```diff
@@ -66,6 +66,8 @@ public ActionResult DeletePostPrompt(int? postID)
             //int index = post.ForumThread.ForumPosts.IndexOf(post);
             var page = GetPostPage(post);
 
+            if (!Global.IsModerator && !((thread.Clan == Global.Account.Clan) && Global.Account.HasClanRight(x => x.RightKickPeople))) return Content("No permission to delete this post");
```
Licho1: use RightEditTexts ?

Licho1:
Does the GUI actually allow clan leader to delete anything? Where is the button?

sprunk:
Same button as regular post delete

Licho1:
Too expensive change, you do fairly complex query for each post displayed.

PR #1663, opened 2017-03-22
test whr on live

DeinFreund:
What do we still need to get this done?


Licho1:
I will do a full review and we can then test. We should also output ram to logs on updates.

sprunk:
> should also output ram to logs on updates.

und dann

>(Licho) will do a full review and we can then test

PR #1691, opened 2017-04-08
StartSetup: Remove suspicious string syntax

KingRaptor:
Because I haven't encountered this syntax anywhere else, MonoDevelop thinks it's an error and neither VS nor GitHub highlight it properly.

May fix PW games not starting.


KingRaptor:
Git blame suggests it's unrelated.

PR #1713, opened 2017-04-14
Add SERVER_GAMEOVER parsing

ashdnazg:
Should be the right solution to the winning team handling.

Untested


ashdnazg:
any way to know the error?

PR #1719, opened 2017-04-18
angle-quote nightwatch github spam, prevents embed

Anarchid:
Discord has a thing that makes urls embedded in angle brackets ( `<` `>`) not embeddable, but still clickable like normal links.

Inflicting this upon Nightwatch's github notifications will make #zkdev@Discord much more readable.


db81:
:heart:

PR #1723, opened 2017-04-23
don't add AIs to fight specs

DeinFreund:
when automagically adding AIs to an empty coop game, specs are counted as players.


DeinFreund:
hm, res shouldn't even contain specs in the first place.

PR #1733, opened 2017-05-04
Fix shutting down

Licho1:


Licho1:
Thank you, this is a fix #1733

PR #1742, opened 2017-05-24
Stop gifv from going out of bounds

DeinFreund:
Upsizes tiny videos though


DeinFreund:
Idk why the previous one didn't compile, but this one is pure css so I'll blame infra if it doesn't either.

PR #1761, opened 2017-07-05
magic unspec command to fix zk's lack of players

DeinFreund:
Firepluk's suggestions make zk great again


sprunk:
Kappa

sprunk:
Afker culture would need to change first (which is hardly doable at all but you can try).

DeinFreund:
sad

Issue #1768, opened 2017-07-10
Prettify MM pausing after mapfeaturing


DeinFreund:
After mapfeaturing, MM actions like joining a queue are no longer echoed to the user(They are updated after the pause). Users should either still be able to join queues or get a quick warning message in the lobby.

Currently it looks like MM is lagging for the pause time.

sprunk:
#1772

PR #1769, opened 2017-07-10
Joining MM just updates status instead of ticking

sprunk:
* ticking might be blocked due to recent mapfeaturing in which case the user would get no echo (#1768)
 * gives some leeway for more people to join the queue in the meantime

### DIFF
--- a/ZkLobbyServer/MatchMaker/MatchMaker.cs
+++ b/ZkLobbyServer/MatchMaker/MatchMaker.cs
@@ -301,9 +301,7 @@ namespace ZkLobbyServer
                     });
 
 
-            // if nobody is invited, we can do tick now to speed up things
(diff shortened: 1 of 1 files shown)


ZkLobbyServer/MatchMaker/MatchMaker.cs
```diff
@@ -301,9 +301,7 @@ private void UpdateQueues()
                     });
 
 
-            // if nobody is invited, we can do tick now to speed up things
-            if (invitationBattles?.Any() != true) OnTick();
-            else await UpdateAllPlayerStatuses(); // else we just send statuses
+            await UpdateAllPlayerStatuses();
```
DeinFreund: This would break if MM is currently not paused. Now players have to wait up to 30 seconds for their battle instead of it being instant.

Issue #1736, opened 2017-05-12
!spec requires more votes than present players


DeinFreund:
`!spec` without any arguments calls a poll to spectate AFK. This requires >50% of unspecced players to agree, thus it's impossible to complete if 50% of players (or just one in a 1v1 room) are afk.

It should probably only count non afk players.

PR #1770, opened 2017-07-10
Don't count AFK players for votes

DeinFreund:
see #1736 

This might have some unintended side effects.


sprunk:
What kind of side effects?

DeinFreund:
This affects all polls, not just spectating, so maybe I didn't think of something.

sprunk:
Why can't the boss handle afkers?

DeinFreund:
This is about votes, those should work independently of the boss. Also in practice the boss is often afk/offline while people are still playing in his room.

DeinFreund:
This is bad in practice. If somebody afks in a room, the joiner can instantly !poll start.

sprunk:
Use the MM if you want to get a notification before a battle.
Host your own room where you are the boss if you don't want to leave yourself to other people's vote.
Don't afk in a public vote room.

PR #1774, opened 2017-07-10
PM people when they get punished

sprunk:
This is useful to give people official warnings. If the punishment has no effect they might miss it completely. Currently they only see the punishment reason if they go to their user page.

### DIFF
--- a/Zero-K.info/Controllers/UsersController.cs
+++ b/Zero-K.info/Controllers/UsersController.cs
@@ -258,7 +258,7 @@ namespace ZeroKWeb.Controllers
 
                 Global.Server.GhostChanSay(GlobalConst.ModeratorChannel, string.Format("New penalty for {0} {1}  ", acc.Name, Url.Action("Detail", "Users", new { id = acc.AccountID }, "http")));
                 Global.Server.GhostChanSay(GlobalConst.ModeratorChannel, string.Format("Reason: {0} ", reason));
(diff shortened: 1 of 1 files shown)


Zero-K.info/Controllers/UsersController.cs
```diff
@@ -258,6 +258,7 @@ public ActionResult LobbyDetail(string id)
 
                 Global.Server.GhostChanSay(GlobalConst.ModeratorChannel, string.Format("New penalty for {0} {1}  ", acc.Name, Url.Action("Detail", "Users", new { id = acc.AccountID }, "http")));
                 Global.Server.GhostChanSay(GlobalConst.ModeratorChannel, string.Format("Reason: {0} ", reason));
+                Global.Server.GhostPm(acc.Name, string.Format("You have been punished: {0}", reason));
```
DeinFreund: "Your account has been penalized:" ?

sprunk: After some thought this should be something that doesn't explicitly say you've been penalized because it might be used for warnings that have no actual penalty effect. How about something along the lines of "Your account has received administrator action: " ?

DeinFreund: I only went with penalty because that's how punishments are expressed on the userprofile, even if they carry no actual right impediments. administrator action sounds better

PR #1779, opened 2017-07-17
Better instructions for replay/etc feature usage

sprunk:
Better wording suggestions welcome


db81:
"You need to be running the game and be logged in there"

PR #1835, opened 2017-10-23
Fix #1834

### LINKED ISSUE #1834: "label" spring exit code -805306369 as hang / killed by user
atm this is printed:

"Spring exit code is: -805306369, assuming crash"

please change to something like this:

"Spring exit code is: -805306369, hanged and killed by user"

"crash" is misleading.

(maybe also see https://springrts.com/mantis/view.php?id=5812)

sprunk:


sprunk:
Looks like PRs arent automatically built anymore?

PR #1839, opened 2017-10-25
Fix #1704, PW influence string rounding.

### LINKED ISSUE #1704: Influence rounding in Battle Instructions
Sometimes the battle instructions will read "23.899999999999999 influence". Round influence to one decimal place because putting it in the string.

sprunk:


sprunk:
Also adds the note about PW structures for defenders which attackers already have.

sprunk:
As a side note it would probably be much better if infra just sent a bunch of numbers and faction names and the actual string was contained ingame (allows localisation, removes the need for infra stable, easier access for GoogleFrog).

Licho1:
It could send sort of string with entities, thats how the string is generated in the first place..

Issue #1838, opened 2017-10-25
itch.io Planetwars


DeinFreund:
It seems planetwars are currently still WIP. 

Would it be possible to do a beta round for the itch.io release? We acknowledge that it isn't completed, but it's playable (and fun imo). Even if there are some things that are still to be implemented right now, testing should give us a better idea of what to work on.

sprunk:
There's quite a lot to work on already: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues?q=is:issue+is:open+label:planetwars

DeinFreund:
Yes, that's why it's still WIP. Does that mean it's unplayable? The main game isn't finished either.

From my experience with the last two PW (which had even more problems), they were still a blast to play. We should clearly mark is as experimental/in development and if people don't like it we can always disable it.

GoogleFrog:
I think it is a bad idea to just run planetwars. I expect that treaties still do not work. In general there are a lot of vestigial mechanics in planetwars which are untested so could be broken. Also is is likely unbalanced. Here is what I would do without any infrastructure support (so all we can do is edit DB and constants):
 * Remove planetbuster.
 * Remove string connector.
 * Test all the treaties and remove the ones that do not work.
 * Set the artefact to cost a lot less energy, perhaps none.
 * Make the energy -> metal conversion much less efficient.

Here is what I would do with some infrastructure support, but not much:
 * Remove planetbuster.
 * Remove string connector.
 * Test all the treaties and fix the broken ones.
 * Add a DB key for reactivation cost. You can instantly pay X metal to reactivate a destroyed structure.
 * Remove automatic energy to metal conversion. This is to make early energy overproduction not a trivial decision.
 * Add a DB key to add metal income to a structure. Possibly use this key to make a mex that costs energy. The key may be useful to have.
 
Ideally we would have an accurate and full list of all the systems and mechanics of planetwars. In order to make good game design we need to know what everything does. Otherwise I feel like removing absolutely everything which is poorly defined in my mind. Create a document detailing the actual current planetwars mechanics if you want to help out in a relatively simple way. I don't think anyone quite knows how planetwars works. Remember the surprising effect of planetbuster.

See: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/1705

sprunk:
It doesn't mean it's unplayable but

>testing should give us a better idea of what to work on

[...]

Issue #1705, opened 2017-04-11
Planetwars mechanics feedback


GoogleFrog:
Here are some game mechanics to consider for next round:
 * Make Artifact ingame destroyable and able to be evacuated. This gives attackers the option of destroying the artifact to prevent VP gain for 10 turns.
 * Reduce ingame structure health. Command center is not easily rushable given its large health pool. Some defense around the CC prevents it from being sniped by a losing team. The vast majority of CC destruction that I have witnessed were achieved by an uncontested, fairly static, force (usually a commpush with turrets). This method of destruction does not care how much health the CC has so we may as well lower it as a buff purely to snipes.
 * Consider shorter evacuation time. Perhaps 10/5 to 8/4?
 * Match colours ingame? Players could receive colours close to their faction colour.
 * String connector: It is cool but subverts many mechanics and makes the map ugly. If we want a drawn out multifront war decided mostly by artifact control then the existence of string connector is bad. It has the potential to make the map feel very small (remove some epicness). Perhaps it is better with warp jammer fix (see https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/1702). It also subverts the defensive metal for distance from HQ mechanic.
 * Consider reducing the energy:metal conversion rate. Currently it is 20 energy for 1 metal, this may be a bit high as it makes the battles less important. Maybe 50 energy for 1 metal?
 * Consider the effect of planet buster and string connector. They bring the game closer to ending in an unstable way. See below.

I dislike planet buster and string connector because they have a permanent effect on the galaxy which is bought about through the actions of few players (the faction leads). Battles should be central to planetwars so I feel like it should focus on battles and fronts, the taking and losing of planets. Every participates in that. The front pushing should also have the opportunity to persist for quite a while, to be solved by diplomacy, coordination and the threat of victory points. Planetwars should be able to support a prolonged struggle otherwise we'll have to run a new one every week.

Here is a way to think about planetwars design: what are the actions that bring the game closers to it's conclusion?
 * Gaining victory points: If 298 victory points are gained then one faction has won.
 * Planet busting: If 69 planets are destroyed then one faction has effectively won.
 * String connecting (sort of). If every planet is connected to every other planet then IP spread will surely overcome the smaller side. More realistically, string connecting cuts off whole swathes of the galaxy from activity.
[...]

PR #1840, opened 2017-10-25
Planet wars: remove E → M conversion

sprunk:


GoogleFrog:
You changed a non-zero value to zero. How will this not break?

sprunk:
Because it's used as `metal = energy * ratio`.

sprunk:
Caveat from https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/1705#issuecomment-293519942

>Dropships were pretty high-powered, though (...) Reducing OD efficiency will increase dropship power.

Licho1:
Why do you want to do this? It is to give more options when building your empire - you can focus on metal farming by just building energy and not energy sinks.

sprunk:
@Licho1, see @GoogleFrog's comment: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/1838#issuecomment-339336844

> Remove automatic energy to metal conversion. This is to make early energy overproduction not a trivial decision.

GoogleFrog:
I claim that E -> M conversion does not really give you more options and leads to unstable economic harassment gameplay. Note that E -> M conversion makes the economy effectively work as follows, with a few caveats:
- Power structures generate metal.
- Most other structures drain metal as upkeep.

Every faction wants to increase its metal income in order to make better infrastructure, especially in a FFA. Therefore they are going to build as much power as they can while building as few other structures as possible. The investment in power structures supports their later infrastructure because they need the extra metal income (energy in the actual game) to pay the upkeep of their infrastructure. If infrastructure had no metal upkeep cost then power structures would be more of a choice.

Factions may also build a little bit of infrastructure to perform quick strikes against the enemy economy or pre-emptive strikes to defend their own, but only if this is efficient (remember it's FFA). Quick, efficient, strikes run counter to the goals of planetwars (to involve all players in a lengthy war) because they are likely to be done by the top players against the weaker players of the other faction. This leads to blaming and excessive timezone optimization instead of teamwork. Ideally there will be relatively important/unimportant battles but they should not outweigh the actions of the rest of the faction. This is also why I dislike superweapons, something really important happened but few people brought it about.

Quick strikes (with warp cores) are effective because they can deal damage to an exponential economy. The exponential economy comes from power generators and everyone has exposed planets covered in power generators. The planets are exposed because there are a lot of planets and the defense structures drain metal upkeep. These dynamics make pre-emptive warp core rushes highly effective. First you can kill their core/dropship production and then you can make economic gains against them. The only thing keeping this in check is FFA dynamics.

Reliance on FFA dynamics to prevent early quick strike rushes is terrible because it will cause a highly unstable game in which one faction may win early. We want the FFA dynamics to regulate the expansion and victory point control of the currently winning faction in the mid to late game. Ideally planetwars would eventually be won by a faction reaching the victory point total. It should be very difficult to entirely halt the gaining of victory points throughout the galaxy, to ensure that the game doesn't stagnate (this can all be balanced through the database).

Anyway, that is why I want to remove E -> M conversion. I want power structures to be relegated to the role of population cap for infrastructure. This population cap only has two purposes in the game mechanics:
- It makes the initial construction of a piece of infrastructure more expensive than rebuilding that infrastructure.
- It lets factions reconfigure their infrastructure without re-paying the cost of the power structure.

Also, as with all structures, power structures are cool to see on planets and protect/destroy. A non-exponential economy will dampen down the effect of the early game and so will not encourage early quick strikes. After thinking about these exponential economies I am now in favour of adding a metal upkeep for ship factories (see https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/1852). I am also considering a dedicated metal production structure because that would be more of a choice. However, while it is worth adding support in the database, I don't think we need to add it to the next round of planetwars.

Licho1:
Ok, lets try this.

Licho1:
Btw don't forget there is still this: PlanetMetalPerTurn = 1;

GoogleFrog:
Planet metal per turn is fine. An income of 1 is relatively small compared to the amounts that power plants were generating and the number of planets is limited. I think it is unusual for a faction to own more than 30 planets and this income would only be half the expected 66 metal income gained from battles per turn. In most cases the faction with the most planets won't have more than 15 more planets than the faction with the least planets so in practise the effect will be quite small.

Issue #1826, opened 2017-10-14
DownloadFileProgress and DownloadFileCancel for wrapper loopback


GoogleFrog:
The UI would like to know the progress of current downloads. It would also like to be able to cancel downloads.

Licho1:
added

GoogleFrog:
DownloadFileProgress should send the file size of the file being downloaded.

Licho1:
Progress and time estimates are not based just on download. For example for pool there are multiple steps. Or file can have dependencies.
Or for engine it also expects time to unzip the file. But it can send other data like dl speed.

Licho1:
done

PR #1850, opened 2017-10-31
Adding chobby wrapper download messages

DeinFreund:
I'd like to help implement #1826 , how can I see loopback messages in Chobby for debugging?

### DIFF
--- a/Shared/PlasmaShared/Utils.cs
+++ b/Shared/PlasmaShared/Utils.cs
@@ -782,6 +782,30 @@ namespace ZkData
         }
 
 
+        /// <summary>
+        /// Converts wait handle into awaitable task with an update callback
+        /// </summary>
+        /// <param name="handle"></param>
+        /// <param name="timeout"></param>
+        /// <returns></returns>
+        public static Task AsTask(this WaitHandle handle, TimeSpan timeout, TimeSpan updateInterval, Action callback)
+        {
+            var tcs = new TaskCompletionSource<object>();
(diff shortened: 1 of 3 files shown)


Shared/PlasmaShared/Utils.cs
```diff
@@ -782,6 +782,30 @@ public static Task AsTask(this WaitHandle handle, TimeSpan timeout)
...
+        {
+            var tcs = new TaskCompletionSource<object>();
+            var expiration = DateTime.UtcNow.Add(updateInterval);
+            var registration = ThreadPool.RegisterWaitForSingleObject(handle, (state, timedOut) =>
+            {
+                var localTcs = (TaskCompletionSource<object>)state;
+                Task.Run(callback);
+                if (timedOut)
+                    if (DateTime.UtcNow > expiration) localTcs.TrySetCanceled();
+                else
```
sprunk: This `else` will bind to the nested `if` but its indentation suggests the intention was otherwise.

DeinFreund: Indeed. That's one more reason why I'd like to know how to test this.

GoogleFrog:
Loopback messages probably appear in the infolog. If they don't then edit the loopback widget locally.

DeinFreund:
Unless you've done any testing, keep in mind that this might not work at all.

Licho1:
You made a PR without any tests at all?

DeinFreund:
Well it compiles and runs, but it might not work as expected. I'll make sure it works later today if you didn't.

PR #1859, opened 2017-11-02
Give donators public exposure

DeinFreund:
They must be praised!


sprunk:
What about privacy? It shows people's real name. I think we shouldn't show it in a public channel (that includes #zkdev) and ideally not show it at all.

DeinFreund:
Oh I thought the name was optional. Yes in this case we shouldn't show the name. We could still show the in game Nick though

sprunk:
Nick sounds fine (especially since you also get a public badger)

PR #1861, opened 2017-11-03
Prevent donations by chanserv

sprunk:


Licho1:
We should simply prevent donations if people are not logged in..

sprunk:
That would fix the chanserv problem, but is it a good idea? I think donating should have as few obstacles as possible. There are people who donate just because something looks cool even before actually having an account.

PR #1868, opened 2017-11-08
Crash reports: better titles

sprunk:


Licho1:
Have you tested this?

Licho1:
this needs to be updated and tested

sprunk:
I'm fairly busy currently, I'll reopen when I get time to test.

PR #1898, opened 2017-11-15
Max players in room: 32 -> 200

DeinFreund:
probably not, but alcur would love it.


sprunk:
regardless of the value I think it should be a GlobalConst

DeinFreund:
If it works we can make it a globalconst. I'd rather not add a globalconst that doesn't actually work and confuses people.

Licho1:
it does work but why do you want to increase it?

DeinFreund:
Alcur said it would improve games.

Licho1:
well first get 200 people then this :-)

PR #1922, opened 2017-12-11
consider ignoring username case on login

DeinFreund:
Misspelled usernames keep leading to confusion because it is expected for usernames to be case insensitive.


sprunk:
As far as I can tell there are some users pairs with same nick differing only by case (legacy from before checks).

PR #1958, opened 2017-12-18
stop Sprung from time traveling

DeinFreund:
How expensive is it to order by time? if it's too expensive we can just ignore the few outliers.


ZkData/Ef/WHR/RatingSystems.cs
```diff
@@ -39,7 +39,7 @@ public static void Init()
                             .Include(x => x.SpringBattlePlayers)
                             .Include(x => x.SpringBattleBots)
                             .AsNoTracking()
-                            .OrderBy(x => x.SpringBattleID))
+                            .OrderBy(x => x.StartTime))
```
Licho1: This field is not indexed, if you do big queries over it, add index for this

DeinFreund:
Sorry, this broke the build again :S
http://zero-k.info:8080/viewLog.html?tab=buildLog&buildTypeId=ZeroK_MasterTestBuild&buildId=9223#_focus=4573

PR #1960, opened 2017-12-19
don't search for specs in battles

DeinFreund:


Licho1:
You might want to seek a game where you watched but ok

PR #1966, opened 2017-12-19
only put people into rating db who played recently

DeinFreund:
This means old people's rating won't be accessible via db queries.


DeinFreund:
`WHR Casual Ratings saved to DB in 0.8555655 seconds`

This really only saves time on the first initialization. Afterwards updates will be applied only to people with an elo change > 10.

PR #1972, opened 2017-12-21
Clearly notify people about unsupported maps

### LINKED ISSUE #1822: Avoid serving unsupported maps
When an unsupported map is selected, there should either be some confirmation required "Do you really want to play on a map that might break the game?" or they should require an explicit command(!map "unsupported map names work in quotes only"). Otherwise people still end up on such maps regularly and encounter bugs/glitches.

In this case people were playing on Zed 2.2 which causes graphical glitches and doesn't render cloaked units.

DeinFreund:
fix #1822


ZkLobbyServer/autohost/Commands/CmdMap.cs
```diff
@@ -20,14 +21,27 @@ public override string Arm(ServerBattle battle, Say e, string arguments = null)
...
+
+
             if (map == null)
             {
                 battle.Respond(e, "Cannot find such map.");
                 return null;
             }
+            else if (!string.IsNullOrEmpty(arguments) && map.MapSupportLevel < MapSupportLevel.Supported)
+            {
+                alternativeMap = MapPicker.FindResources(ResourceType.Map, arguments, MapSupportLevel.Supported).FirstOrDefault();
```
Licho1: Dont do this. I dont buy this supported/unsupported BS. 
This will restrict what you can easilly pick.
Instead maps should be marked as "duplicate of" and "isspecial" I would like to remove IsSupported completely.

DeinFreund: This won't stop you from picking any maps, it'll just suggest you to choose a supported one instead. We have so many maps that were made for other mods and don't work at all with zk. I don't think we can blacklist all of them.

Anarchid: The suggestion seems pushy. Notifying people is good, telling them what to do is not.

DeinFreund: There have been multiple bug reports because of people playing on a broken map. If we could disallow all of these known broken maps that would work as well, but I doubt we have the resources to identify each of them.

DeinFreund:
This will only do suggestions, `!map` will still behave the same. How should I word the suggestions to be less pushy @Anarchid ?

Anarchid:
I have re-read the code and i don't think it's that pushy now. 

The pushy part is where you suggest a possibly unrelated but supported map instead of what the users clearly wanted with their ill-gotten !map command. But because the same arguments are passed to the alt map selector, there's also a reasonable chance that they do get something that's relevant, like a newer version of red.

DeinFreund:
That was the idea, often people accidentally select an old version of a map. This won't stop you from doing that, but it'll remind you there's a different, recommended version available.

PR #1987, opened 2018-01-02
balancer tries to reduce variance

DeinFreund:


Licho1:
This has been tested with poor IRL results.  Average is the best predictor we made so far for determining team strength.
Variance can only be optimized if averages stay the same.

DeinFreund:
I would probably go for some very small factor for the stdev, meaning it'll choose the best stdev once elo is already matched.

PR #2007, opened 2018-01-12
Autospec

### LINKED ISSUE #1832: Consider auto-spec of AFK players at !start
This request shows up from time to time, e.g. [here](http://zero-k.info/Forum/Post/174861#174861) and today in #zk channel. 

For optimal results, AFK should detect out-of-lobby status, e.g. someone who is using web browser at that moment should not be considered AFK. Also fix !ring and make Springie use it on AFKers when a !start vote occurs.

There's a chance that AFK people will return when the game starts, but if more than one of them fail to do so (especially if they're on the same team) I expect people will get annoyed in pretty short order. I think it's not unreasonable to expect people who want to play to be actually present.

DeinFreund:
fix #1832

### DIFF
--- a/Zero-K.info/Views/Home/HomeIndex.cshtml
+++ b/Zero-K.info/Views/Home/HomeIndex.cshtml
@@ -159,7 +159,7 @@
     else
     {
         <div id="unit" class="border relative text-left">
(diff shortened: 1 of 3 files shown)


Licho1:
AFK is unreliable! Do not do it!

Licho1:
As in first at least partially fix AFK by providing system-wide AFK status to lobby.

PR #2008, opened 2018-01-12
[wip] chobby offline mode

DeinFreund:
I've added caching of chobby version and engine version for offline play.

### DIFF
--- a/Shared/PlasmaShared/GlobalConst.cs
+++ b/Shared/PlasmaShared/GlobalConst.cs
@@ -34,6 +34,7 @@ namespace ZkData
 #else
             Mode = ModeType.Local;
 #endif
+            Mode = ModeType.Live;
         }
 
         static void SetMode(ModeType newMode)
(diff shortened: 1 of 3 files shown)


Licho1:
I think it is dangerous...
For steam, you want to silently run even if nothing online works...

Shared/PlasmaShared/GlobalConst.cs
```diff
@@ -34,6 +34,7 @@ static GlobalConst()
 #else
             Mode = ModeType.Local;
 #endif
+            Mode = ModeType.Live;
```
Licho1: You shall not pass! This will make test version of the game screw the live databas!

DeinFreund: this is just so i can test locally and why i wrote WIP so you don't merge anything

DeinFreund:
Even though Chobby worked for me just fine offline. If it turns out it doesn't this is what I'd use to make it work. Maybe the message could be modified to "Unable to establish connection to server, some features will not work. Try to run the game anyway?". This assures that chobby doesn't silently start offline and then breaks without warning due to some missing download.

PR #2010, opened 2018-01-12
allow commands to be used independently of chat location, don't broad…

DeinFreund:
…cast commands to lobby

fix admins not being able to run commands in MM

### DIFF
--- a/ZkLobbyServer/ServerBattle.cs
+++ b/ZkLobbyServer/ServerBattle.cs
@@ -704,7 +704,7 @@ namespace ZkLobbyServer
             spring.DedicatedServerExited += DedicatedServerExited;
 
             spring.DedicatedServerStarted += DedicatedServerStarted;
-            spring.PlayerSaid += async (s, e) => await spring_PlayerSaid(s,e);
+            spring.PlayerSaid += spring_PlayerSaid;
             spring.BattleStarted += spring_BattleStarted;
         }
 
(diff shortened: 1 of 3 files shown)


ZkLobbyServer/ServerBattle.cs
```diff
@@ -695,7 +704,7 @@ private void SetupSpring()
             spring.DedicatedServerExited += DedicatedServerExited;
 
             spring.DedicatedServerStarted += DedicatedServerStarted;
-            spring.PlayerSaid += spring_PlayerSaid;
+            spring.PlayerSaid += async (s, e) => await spring_PlayerSaid(s,e);
```
Licho1: This might be dangerous thing to make.
Make sure that 'UnsubscribeAllEvents' is applied to this.
This event listener must be removed, or garbage collector wont dispose spring class properly, possibly leading to memory leak.

Licho1: I would recommend keeping the vent non-async.

Licho1:
Has this been tested?

DeinFreund:
No.

Licho1:
why do you close it? Its not so hard to test this is it?

DeinFreund:
It's not, but I still don't have time to test and debug right now so I wanted to make sure it's not mistaken for active.

Licho1:
oh ok closing until tested then

DeinFreund:
Reopening in case testing it on test server could be considered now.

DeinFreund:
Current bug:

Admins can't administrate MM battles. If someone is behaving badly in a MM match they have to ban him, but he can still continue playing and doing bad things until the battle ends. It's nearly impossible to kick someone from an MM battle, because the dedicated server continues running even through a server restart.

PR #2011, opened 2018-01-12
Battleautoquit

### LINKED ISSUE #1648: Battle not recorded
I was just playing a MM game which didn't have post battle discussion and wasn't recorded in the battle db. 

Timestamp: (GMT+1) 20170316_203712_Fairyland v1

I rejoined once during the game.

DeinFreund:
fix #1648


ZkLobbyServer/ServerBattle.cs
```diff
@@ -275,6 +283,11 @@ public bool IsKicked(string name)
             var pwd = GenerateClientScriptPassword(conus.Name);
             spring.AddUser(conus.Name, pwd, conus.User);
 
+            if (spring.Context.LobbyStartContext.Players.Any(x => x.Name == conus.Name) && conus.MyBattle != this)
```
Licho1: This is really odd place to put this..
the server KNOWS when user rejoins. It has to send him password too.. there is talker/spring event, should not be hacked here

ZkLobbyServer/ServerBattle.cs
```diff
@@ -590,6 +603,7 @@ public virtual void ValidateBattleStatus(UserBattleStatus ubs)
                     });
 
             toNotify.Clear();
+            CheckCloseBattle();
```
Licho1: Should not be hacked in say

DeinFreund:
Not tested but IMO a rather major bug.

DeinFreund:
Here are two scenarios that are fixed:
User A and B are playing 1v1 MM. User A has his internet drop out or the game crash and has to disconnect. He rejoins the battle and continues playing. User B is doing badly so he decides to leave. The battle is killed because there are no more users inside, User A doesn't get his victory.

User A and B are playing 1v1 MM. User A has his internet drop out or the game crash and has to disconnect. He rejoins the battle and continues playing. User B now has to rejoin the battle as well due to some issue or other reason. The battle is considered empty and exited even though both users still want to continue playing and User A is still in-game.

PR #2069, opened 2018-03-03
Set preferred rating system in Battle, validate in RatingSystems. Isolate rating systems from each other.

DeinFreund:


GoogleFrog:
Can't we manually tick the games if counting them for rating makes sense?

DeinFreund:
Of course you can do everything manually, but I doubt you're going to actually do it in the end.

DeinFreund:
Nearly a month has passed since the last tourney and nobody has marked the battles.

PR #2109, opened 2018-03-24
Master -> stable

KingRaptor:
Planets are actually clickable now (real sorry about this one!)


KingRaptor:
Site hotpatched, PR no longer needed for now

PR #2130, opened 2018-03-25
only delete clan polls on clan leave

### LINKED ISSUE #2129: Leaving clan cancels Election
DeinLobster was being elected for Emprah, but then he created and left an Empire-Clan. After leaving, even though he remained in the Empire faction, his election for Emprah was cancelled.

DeinFreund:
fix #2129

@Licho1  does this have unintended side effects?


Licho1:
probably no its safe

PR #2135, opened 2018-03-26
add PlanetWarsMaximumIP, fix #2134

### LINKED ISSUE #2134: Add max IP to global constants
The maximum IP on a planet should be a constant so we can easily scale up PW if required. With more players it would make sense to make them take more time to conquer a planet. I prefer scaling the maximum IP on a planet to scaling the sources of IP gains because we're already at the first decimal place with IP and I don't want to go any deeper.

DeinFreund:


GoogleFrog:
I'm noting that I think it is correct to not change InfluenceToLosePlanet. Planets are rarely lost this way and I don't want to have to remember to change the HQ DB entry that prevents IP decay below 11.

PR #2142, opened 2018-03-28
use precached factions

sprunk:


Licho1:
Precached faction of.GetTreatyRights wont work if the context on which they were created is no longer valid.

sprunk:
The context on which `of` is created also goes out of scope before the list is returned, I thought it might be relevant.

Licho1:
It might be! It should stay in scope, if it goes out of scope it might crash or otherwise fail if context is disposed.

PR #2145, opened 2018-03-30
stable pw lagfix

DeinFreund:


GoogleFrog:
I don't think PW games should have a rating or have any effect on rating.

PR #2163, opened 2018-04-06
disallow trading HQ planets

DeinFreund:


GoogleFrog:
This is not what IsVictoryPlanet is for.

GoogleFrog:
This is just wrong. EffectIsVictoryPlanet is 1 on artifact planets because they originally caused victory. Now I would use that same key to show who owns artifact planets in the faction UIs. The key you actually want for this PR is OwnerChangeWinsGame.

sprunk:
fe60ff506bf12f19512320690bec86581a4be659

PR #2172, opened 2018-04-08
Unowned structure production fix

KingRaptor:
Currently structures without an owning account (or any where the owner somehow left faction without abandoning the structures) do not produce any dropships, bombers or warp cores. This PR fixes that.


Licho1:
This is intentional. Planet owner or leader can "confiscate them".

KingRaptor:
I feel it's undesirable for production to be wasted if faction leaders/planet governor fail to notice a structure is unowned, and I can't currently think of a benefit from allowing wastage.

GoogleFrog:
What is the purpose of this intention?

Licho1:
PW has had many different rules and structures should produce for their masters.
If we wish structures to always have owner, we should give them owner and not reroute production

Licho1:
Or sell it

PR #2175, opened 2018-04-08
Give helpful reasons if commands cannot be run. …

DeinFreund:
This especially addresses cases where users would switch to the lobby view and try to run non-ingame commands while the game is still running.

### DIFF
--- a/ZkLobbyServer/autohost/Commands/BattleCommand.cs
+++ b/ZkLobbyServer/autohost/Commands/BattleCommand.cs
@@ -82,8 +82,9 @@ namespace ZkLobbyServer
         /// <param name="battle"></param>
         /// <param name="userName"></param>
         /// <returns></returns>
-        public virtual RunPermission GetRunPermissions(ServerBattle battle, string userName)
+        public virtual RunPermission GetRunPermissions(ServerBattle battle, string userName, out string reason)
         {
+            reason = "";
             if (Access == AccessType.NoCheck) return RunPermission.Run;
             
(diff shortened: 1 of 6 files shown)


ZkLobbyServer/autohost/Commands/BattleCommand.cs
```diff
@@ -120,14 +121,20 @@ public virtual RunPermission GetRunPermissions(ServerBattle battle, string userN
             }
 
             var defPerm = hasAdminRights ? RunPermission.Run : (isSpectator || isAway ? RunPermission.None : RunPermission.Vote);
+            reason = "This command can't be executed by spectators. Join the game to use this command.";
             if (defPerm == RunPermission.None) return defPerm;
             if (defPerm == RunPermission.Vote && count<=1) defPerm = RunPermission.Run;
 
+            reason = "";
             if (Access == AccessType.Anywhere) return defPerm;
             if (Access == AccessType.Ingame && s.IsRunning) return defPerm;
```
Licho1: It should be refactored a bit for clarity...
perhaps something like


if (Access = Ingame) {

if (isrunning) return defPerm;
else { 
  reason = "blah"
return none;
}

DeinFreund: I did that first but thought it was so long that it's hard to read. I'll do it that way if you prefer it though.

DeinFreund:
Tested during a game on test server.

PR #2180, opened 2018-04-10
Fix #2174, expose PW info to specs

### LINKED ISSUE #2174: Send PW information to specs as well
The PW information panel that appears for attackers and defenders could also have a spectator version. This would let the spectators know the stakes in the battle and make it more interesting.

sprunk:


Licho1:
For consistency i think if site generates full texts for non specs, it should do the same for specs.
Otherwise its place jumping when you edit it and you easilly miss the change

sprunk:
My plan is to implement the texts ingame, then remove the full text generation from site. Now it is kept for the transition period.

sprunk:
Also it can't generate it for specs because specs can't have `playerKeys` (they join after script is generated)

GoogleFrog:
I want a namespace for these new modoptions, perhaps "planetwars_". Otherwise we'll have all these generic reserved modoptions and may run into trouble.

Licho1:
Texts in game might require even more data to be sent btw

Licho1:
You are not sending attacker, defender facitons etc..
I suggest generating texts on server for now. Then replacing it all in one go. It will also mean one go.

sprunk:
Everything is exposed. Factions were already being sent:

https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/c45c8e10fd15489e1e0dc3a60dbf7b1b0a14cf18/ZkLobbyServer/SpringieInterface/StartSetup.cs#L71-L85

https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/c45c8e10fd15489e1e0dc3a60dbf7b1b0a14cf18/ZkLobbyServer/SpringieInterface/StartSetup.cs#L113

sprunk:
It's not possible to fully add texts for specs serverside because they are not necessarily even known to server (what if someone watches a replay?)

Licho1:
I meant generate on server, store texts in modoptions instead of values.
Then it will work in replay too.

PR #2198, opened 2018-04-17
Playerkey for elo ordering

sprunk:


Licho1:
what is it for if you have elo?

sprunk:
From time to time people propose to hide elo in at least some modes. This never gets anywhere because ingame relies on elo (for lagmonitor priority). If ingame knows ordering then it won't need elo and so if we wanted to experiment with hiding it could be done without hassle.

sprunk:
Basically this https://github.com/ZeroK-RTS/Zero-K-Infrastructure/commit/6757a908218d6b0a4a33495626e30c837813e463#diff-80d19dd0b2f66e953ffb3059d6714ddfR93

It was later removed because it was unused at the time but I think restoring it and making ingame use it gives us greater flexibility later on.

PR #2204, opened 2018-04-20
Fix ownerless planet income

sprunk:


sprunk:
The proper solution would be to make sure planets are always owned but I think the "seed" planets are there right after reset before anybody rejoins the factions.

ZkData/Ef/Galaxy.cs
```diff
@@ -94,9 +94,12 @@ public void ProcessProduction()
...
 
             // planets generate metal
-            foreach (var p in Planets.Where(x => x.Faction != null && x.Account != null))
+            foreach (var p in Planets.Where(x => x.Faction != null))
             {
-                p.Account.ProduceMetal(GlobalConst.PlanetMetalPerTurn);
+                if (p.Account != null)
+                    p.Account.ProduceMetal(GlobalConst.PlanetMetalPerTurn);
+                else
+                    p.Faction.Metal += GlobalConst.PlanetMetalPerTurn;
```
Licho1: p.Faction.ProduceMetal should be used for future proofing

sprunk:
Done

Issue #1741, opened 2017-05-21
News window is small and ugly


GoogleFrog:
The news window is too small. It should at least be wide enough to not have a horizontal scrollbar and to show the entire image. Ideally it would also show the introduction of the news. To be concrete, it should show 10 lines of text. This is enough for short posts and the introductions of large ones.

![uglynews](https://cloud.githubusercontent.com/assets/1318249/26280901/db1bb626-3e28-11e7-98e1-f660774191ea.jpg)

Licho1:
It was made smaller because people complained its too big..

GoogleFrog:
Those people were wrong. At a bare minimum it should not have a horizontal scrollbar and should show the image as well as the first paragraph of text.

maackey:
IMO it should not be a pop-up at all. Login interrupts are like rotting fish in your air ducts. Sure it gets your attention, but not in a nice way. 

If a news update is important enough to be an in-your-face alert, at the very least it should be able to be closed effortlessly (like the modals in the new admin actions -- clicking anywhere outside popup will close it). On smaller screen sizes the news "alert" would expand to be full width to give room for the content, but that also means it's more difficult to close it 😒 

The alternative solution I was thinking of just adding news/headlines into their own modular section on the home page. It'll never be too large or prevent them from doing what they actually came to the site to do. Depending on how much control the users should have, placing sections the "headline" could always be placed first / on top -- and disappear when closed, available elsewhere in "news", or perhaps just group the headline into "news" in the first place and let the user put it wherever they want, depending on how important news updates are to them. I dunno, I haven't played around with this part a lot yet. 

Regardless of which method of implementation -- this is a pretty straightforward styling issue for the most part. Image header should probably always be set to 100% width to prevent horizontal overflow, and also have a max-height to allow the first paragraph to have enough room. However, on smaller devices this may require cropping out significant portions of the image, which may lead to some unintended consequences (eg. a boring / action out of frame header image).

jquery-ui has been annoyingly intrusive in other parts of the site, but this should be relatively easy to completely rip out and replace it with something better.

PR #2219, opened 2018-04-26
use db ratings for whr while initializing

DeinFreund:
Tested to work locally. Not sure if this is the intended solution though.


sprunk:
#1741

DeinFreund:
Let's make it bigger because people complained it's too small! (Also screens have increased in resolution since 2007)

PR #2224, opened 2018-04-28
Increase MM maximum expansion time to 3 minutes

DeinFreund:
I think 60 seconds was just set for testing originally? I like the fast matching, but with our increasing steam activity I feel we could make use of better quality matches.


Licho1:
thoughts?

DeinFreund:
At least for 1v1, it seems load times are many times longer than wait times now. 30 secs of wait time would be perfectly acceptable. (Although most newbies seem to be matched at initial range anyway)

GoogleFrog:
I really want to be able to change the numbers 50, 350 and 180 without any infra restart.

PR #2235, opened 2018-04-30
 make matchmaker synchronous

DeinFreund:
Any method that calls onTick has to wait for ontick to finish first.


Licho1:
This change appears to be pretty big .. or perhaps formatting change make it look big.

DeinFreund:
Requires testing either way. If you have suggestions let me know.

Issue #2234, opened 2018-04-30
Fix MM ban properly


GoogleFrog:
The MM can grind to a halt if someone, for example, wants to play 3v3 and keeps rejecting 2v2. The ban should be fixed for sanity purposes.

Licho1:
We can also consider more queues to avoid this issue.  2v2 and separate 3v3-5v5

PR #2237, opened 2018-04-30
splitting queues and making things more readable

Rekojeht:
split queues as suggested here https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/2234

and made BannedSeconds more readable

ref: Issue #2234

### DIFF
--- a/Fixer/NubSimulator.cs
+++ b/Fixer/NubSimulator.cs
@@ -18,14 +18,16 @@ namespace Fixer
         public void RunNub(int num)
         {
             var tas = new TasClient("Nubotron");
+
+            var maps = AutoRegistrator.RegistratorRes.campaignMaps.Split('\n');
             var name = "TestNub" + num;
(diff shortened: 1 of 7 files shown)


DeinFreund:
Code looks good, but I'm unsure whether we want to add so many queues and use the same map pool for 2v2 and 5v5.

sprunk:
Yeah if you're adding 5v5 then the map pool should be reconsidered because it's starting to get large. IMO just split into 2 3 4 for now.

MikeHolder:
Okay, I removed 5v5. I think Licho mentioned it so that's why I had it there. :)

GoogleFrog:
Splitting queues has the dangerous side-effect of making it harder to find games if people sit in individual queues. The UI needs to push people towards joining multiple queues at once. Also I forget whether the space scales automatically or needs a lobby update.

Anyway, looks like a decent change.

PR #2249, opened 2018-05-01
remove accept requests from mm

DeinFreund:
Just start whatever battle is possible. AFK players need to be removed from queue by lobby.

### DIFF
--- a/ZkLobbyServer/MatchMaker/MatchMaker.cs
+++ b/ZkLobbyServer/MatchMaker/MatchMaker.cs
@@ -330,18 +330,8 @@ namespace ZkLobbyServer
 
 
             // if nobody is invited, we can do tick now to speed up things
-            bool doUpdates = false;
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I'm highly against this. Removing accept is terrible because you'll end up loading battles which don't occur.

DeinFreund:
The idea is to move the accept part to joining the queue. Assuming queue wait times are 2 minutes at most, this should be fairly equal to accepting the game itself.

The hard part is making sure people are still active when queue wait times exceed those 2 minutes. For this, the lobby could ask players if they want to remain in the queue if it detects afk.

sprunk:
>the lobby could ask players if they want to remain in the queue if it detects afk

This is what accept does, except doesn't boot out people prematurely.

>Assuming queue wait times are 2 minutes at most

Pretty bold assumption.

sprunk:
Superseded by #2258 (by setting interval = 9001)

PR #2254, opened 2018-05-03
add minimum win chance to mm parameters

DeinFreund:


Licho1:
We should improve which players are added to proposed battles. As in improve ordering of who gets added to battle.

DeinFreund:
I think the current principle of using players that are closest by Elo is pretty good. Either way this is just a safeguard against bad battles. The algorithm can be improved independently.

GoogleFrog:
Wow is the win chance limit all implemented? This would be great to get tested and deployed soon as it would really fix teams MM. Currently I fear clicking the teams button because in the past it served me stomps.

PR #2258, opened 2018-05-04
players dont need to accept twice within an interval. This should mak…

DeinFreund:
…e MM less annoying and remove the risky cases where you have to accept twice in a row.

I think accepting twice in a row may lead to race conditions so removing this should make MM less risky. People are also very annoyed at having to accept many battles without any actually starting. This makes sure you can't join a queue and instantly decline.


DeinFreund:
Was a suggestion by sprung I think. The idea is not to ask players if they're ready over and over again. If you pressed ready or recently joined a queue just give them a game. If someone hasn't clicked any mm buttons recently, ask for ready to make sure they're still there.

sprunk:
This is a lite version of #2249 that can emulate the current system of always requiring accept (by setting the interval to 0) or the system from #2249 of never needing accept (by setting the interval to 9001).

I am ambivalent about actually setting it to a >0 non-infinite value:
```diff
+ surely if someone was ready 10s ago they still are so the AFK check is not hindered.
+ according to DF, having to click the accept button is a major hurdle to match creation.
- opaque
- if someone was ready 10s ago surely they can click again.
```

DeinFreund:
People actually intentionally don't click ready at times. I don't see why we should even allow this.

DeinFreund:
Right now it seems we're more on the route towards reintroducing the decline button, while this PR was aiming to remove that possibility and to automate accept.

PR #2297, opened 2018-05-19
Improve report page

sprunk:
* different text for contacting vs reporting
 * emphasize (b+u) the need to provide evidence because so far 90% of new people's reports dont contain it


KingRaptor:
Copied .cshtml page manually to live server as well.

PR #2366, opened 2018-07-17
MM party elo matching improvement

Angeliyah:
The range limitation for party is reducing game for party directly. We dont want to suppress party by deducting their games, but to encourage them do party against stronger enemy. The rank dynamic config thingy is still used to determine party elo range. But this time not to shorten or widen the range, its here to determine how the range will be shifted. 100% deduction for 1:1 enemy, 0% limitation for 2x stronger enemy.


ZkLobbyServer/MatchMaker/MatchMaker.ProposedBattle.cs
```diff
@@ -88,13 +88,13 @@ public bool CanBeAdded(PlayerEntry other, List<PlayerEntry> allPlayers, bool ign
                 }
 
                 var width = owner.EloWidth * widthMultiplier;
-                if (hasParty)
-                    width = width * DynamicConfig.Instance.MmWidthReductionForParties;
+            /*    if (hasParty)
```
sprunk: Don't comment stuff out, remove.

ZkLobbyServer/MatchMaker/MatchMaker.ProposedBattle.cs
```diff
@@ -130,12 +130,12 @@ private int GetPlayerMaxElo(PlayerEntry entry)
...
             private int GetPartyMaxElo(PartyManager.Party party, List<PlayerEntry> players)
             {
-                return (int)Math.Round(players.Where(x => x.Party == party).Select(GetPlayerMaxElo).Average());
+                return (int)Math.Round(players.Where(x => x.Party == party).Select(GetPlayerMaxElo).Average()*(2-DynamicConfig.Instance.MmWidthReductionForParties));
             }
 
             private int GetPartyMinElo(PartyManager.Party party, List<PlayerEntry> players)
             {
-                return (int)Math.Round(players.Where(x => x.Party == party).Select(GetPlayerMinElo).Average());
+                return (int)Math.Round(players.Where(x => x.Party == party).Select(GetPlayerMinElo).Average()*(2-DynamicConfig.Instance.MmWidthReductionForParties));
```
sprunk: You cannot multiply the rating, you can only multiply the gamma.

Also, `2-x` makes for a very unintuitive value.

Angeliyah: 12:53] Archangel: when its 0% threshold, ppl in party fight against 2x more powerful players, corresponding to the original "no gem at all for party" situation
[12:53] Archangel: when its 100% threshold, parties will be fighting with enemy at their own exact litral level
[12:54] Archangel: corresponding to the original "newbie bullier" situation

Angeliyah: but it would be good if we can apply the multi to the player strength. But how can one accessthe player strenth in this scope..? or we can just do an anti log to restore the original strength from elo, apply the multi, and log it back.

sprunk: If you want to make people fight 2x more powerful opponents then add 120 rating (== `400 * log10(2)`). At this point you're operating on the familiar ratings so make the DB variable just mean this offset (else it's unintuitive).

ZkLobbyServer/MatchMaker/MatchMaker.ProposedBattle.cs
```diff
@@ -144,7 +141,7 @@ private int GetPlayerMinElo(PlayerEntry entry)
                 return (int)Math.Round(CutOffFunc(entry.MinConsideredElo));
             }
 
-            private bool VerifyPartySizeFits(PartyManager.Party party)
```
sprunk: random whitespace screwup

Angeliyah: :<

sprunk:
Nothing seems to have happened.

PR #2368, opened 2018-07-20
Implemented fast balance algorithm for teams

DeinFreund:
This opens up an interface to implement balance checking for proposed MM matches without overloading the server.

The balance algorithm now runs in O(2^(n/2)) compared to O(n!/(n/2)!^2).


Licho1:
Do you please have a link to algo description?

Issue #2245, opened 2018-05-01
Team MM creates some wrong battles


GoogleFrog:
This battle was created. I suspect the Russians were in a party.

http://zero-k.info/Battles/Detail/504741?

![image](https://user-images.githubusercontent.com/1318249/39473326-a7f728c4-4d91-11e8-93ce-048da4d68de8.png)

GoogleFrog:
![image](https://user-images.githubusercontent.com/1318249/39475040-f2d79b5a-4d99-11e8-8af6-11b56c9a3810.png)

GoogleFrog:
http://zero-k.info/Battles/Detail/503894?ShowWinners=True

The tree lowest rated players in this game were on one team, and Sigero was on the other. The mumble clan may have done VC but then why was Kcodarc there instead of Nippahh?

Lasica:
More examples:
http://zero-k.info/Battles/Detail/515477
http://zero-k.info/Battles/Detail/515407
http://zero-k.info/Battles/Detail/515446
http://zero-k.info/Battles/Detail/515477

Sigero and zenfur were in a party.

Issue #2321, opened 2018-06-01
Matchmaker is allowing completely imbalanced matches


gajop:
This happens in 2v2:
- Me and Corvus_Corax matching against Zolfy and Sushino, the "predicted chance to win" is 80%+

This also happens in 1v1:
- Me matching against newbs
- Me matching against Drone

All of these are pointless waste of time... Make matchmaking stricter, we have the playerbase for it.

gajop:
PS: The Zolfy + Sushino balance happens maybe because they are in the same team (queuing together)

sprunk:
Related #2246

DeinFreund:
We should try to minimize the time players spend in the lobby, especially newbies. Thus every button should focus on facilitating playing the game. A "match maker" that doesn't find a match within a newbie's attention span sounds like a trap barring them from the game.

If we can't offer balanced matches that's rather unfortunate, but still better than a game you can't play.

sprunk:
As far as I can tell, currently the MM is set to 550 elo width and 0.55 party multiplier which results in up to 96% win chance in randoms vs randoms, 85% in party vs randoms and 72% in party vs party.

Issue #2246, opened 2018-05-01
Team MM prediction sanity check


GoogleFrog:
Team MM should do the game balance for the expected game (if everyone accepts, lets keep it simple) and then reject it if the predicted win chance of one side is above 70% (ideally non-restart configurable).

See https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/2245

PR #2372, opened 2018-07-26
reintroduce minimum win chance for MM team games. 

DeinFreund:
This time without expensive balancing (No more calls to DB)

When applying this change, MmWidthReductionForParties should be reset to 1 (No width reduction). When active, this will replace the existing width reduction.


DeinFreund:
@GoogleFrog could you confirm that you'll do the configuration once this would be live? I don't have access to that and it'd be counterproductive to have this check without proper configuration.

sprunk:
@Anarchid or @Histidine91 could also put the proper value (or if that's the only merge blocker you can just put 0.3 as the default since it's pretty sensible anyway).

Also, as far as I can tell this fixes #2245, #2246, and #2321.

PR #2373, opened 2018-07-27
chobbyla sends idletime to chobby

DeinFreund:


DeinFreund:
It's the code from ZKL

sprunk:
looks good

PR #2384, opened 2018-08-03
Master -> stable

Histidine91:
Autohost improvements; fix Steam not building


DeinFreund:
Autohost is tested, but now is a bad time for stabling. I wouldn't do it before monday morning.

Issue #2385, opened 2018-08-04
Future timestamps in Contributions


sprunk:
Some timestamps are from the future:

![contribs](https://user-images.githubusercontent.com/2573076/43678763-955bee34-9819-11e8-9f04-beff9bfeb757.png)

DeinFreund:
What time zone are contributions stored in? UTC?

PR #2390, opened 2018-08-08
handle donation timestamps as local time. workaround for #2385

DeinFreund:


DeinFreund:
Teamcity doesn't wanna give me green light :(

PR #2395, opened 2018-08-13
optionally join all players into MM once a battle ends

DeinFreund:


Licho1:
What is this? Is htere a MM queue 16-24? Or is it for when you have too few people like 3 ?

DeinFreund:
This is a way to prove or disprove Sprung's theory that people in CF would play MM if it was populated. The idea is to ask everyone in big rooms whether they want to play a small teams MM after a game. It will try to make a game with those who accept and remove everyone from MM once the big game gets going again. If somebody is desinterested (declines once) it will not ask him again for the rest of the playing session.

It is implemented using the standard accept/decline for MM, but the texts will be changed chobby-side to make sure people understand what's happening. In order to guarantee quick yet good MM games, the users start at maximum width and win chance limiting is used.

If there is truly a crowd of people who'd play small teams in the big rooms, this should allow for some more MM games. At the same time you don't have to sit out the clusterfuck just because you want to play small teams, i.e. the players get a game either way.

Licho1:
I think people who are in CF want to play CF..  oh well..

DeinFreund:
Probably you're right, but at least this gives us a non-destructive way to
test that.

GoogleFrog:
What exactly is going to happen, on the user end? A detailed look at a naive design reveals it to be bad, so I'd feel better if the design was exactly stated. Here is the naive design:
 - Players have the tag "post-CF-MM" by default.
 - Upon finishing a CF, all players in the CF with the "post-CF-MM" are asked whether they want to join teams MM.
 - Players that click Yes join the teams MM queue.
 - Players that click No have their "post-CF-MM" removed.
 - Players are removed from the MM queue when a game starts (I forget if this is already the case for players).

What sort of stable environment is this system likely to achieve? I see two equilibrium.
 - Players click 'No' every time. They do this because they pressed 'Yes' in the past and nothing happened, beyond the occurrence of the annoying popup (remember that all popups are annoying.
 - Enough players click 'Yes' to maintain trust in responding 'Yes'. This is a precarious equilibrium because people are much less likely switch from a Noer to a Yeser than the other way around.

It follows that you need to limit your presentation of the popup to situations in which there is a good chance of starting a MM game. 

Hang on....

GoogleFrog:
I glanced at the code for some details and it seems like you just join people to MM automatically? So now you're violating peoples autonomy and spamming everyone in the MM queue with multiple join requests because some sizeable fraction of the CF players will click decline.

We first need to take into account that we are not putting everyone in the CF into the MM queue. We are just adding the fresh CF players, the ones that just finished their first game of the session, because the other players previously clicked decline on a match notification. They clicked decline because either:
 - They want to play a CF instead. Now is the best time to play CF because the CF game just ended.
 - They clicked accept last X times but the match was rejected. Decline frees them from the constant barrage of useless popups.

What is the rate at which fresh players join the CF room? This is approximately your MM pool. I still feel like the equilibrium of this system will rest at the CF players clicking decline. Perhaps a game could have been created in some cases but it will be impossible for the players to coordinate this, because the system will train them to click decline. Now, effectively, we have annoyed the CF players with popups and poisoned the teams MM pool for anyone else who would attempt to use it.

The players are not just being trained to decline in order to avoid popups, they are also being trained to avoid missing the CF game. Players **will** click accept without knowing what it entails, this will lead them to abandon their MM game. This puts the other players, those who wanted to play the MM game, in the position of missing the CF game because it took them too long to figure out that their MM game was dead. This cannot be ignored and there is no helpful text box that will fix the issue.

Basically, I think any such system needs to be backed up by estimates on which sorts of players are likely to result in a successful MM popup, and then only join people in those situations. These should be conservative estimates, as otherwise we train most people to click decline. Only those who are super keen to play small games, given the risks, and perhaps know a bit about the system will remain.

GoogleFrog:
If we're catering to the knowledgeable powerusers how about the following system?
 - Automatically remove players (not spectators) from MM when a battleroom battle starts. This probably requires extra information from the server.
 - Automatically re-add the removed players to MM when their battleroom battle ends.
 - Tell the powerusers that this system exists, and have them use and spread the system to look for MM while waiting in a CF. If they don't use the system then they didn't want to play MM anyway.
 - Possibly add a setting to automatically start MM on some queues when the lobby launches. The powerusers can set it to automatically run in the background.

The big big issue with this proposal is that I don't see what to do when the "MM accept/decline" window appears, shortly followed by the CF battle starting. At this point the respondent does not know whether the MM game will be created, but if they don't join the CF battle they will miss out on the battle. If they auto-decline the MM and join the CF battle then we've poisoned the MM pool, making it worse for everyone. If they join the CF and then potentially join the MM game we've broken the team balance in the CF room, potentially forcing a restart and much wasted time.

I expect this situation to be common as we're potentially dumping a few new MM players into the pool for a few minutes before the next CF game starts. However, a solution to this problem must be part of any CF -> MM system (such as this PR) so my alternate system can just use whatever solution is in this PR.

DeinFreund:
Currently, it will automatically join players into MM after each battle in "QuickPlay" mode. Whether players are joined into MM is determined by the amount of people playing in the room. Seeing that some of them might have denied MM previously and thus be ineligible for this, this should be changed to look at the amount of people eligible for MM. We will have to determine ourselves, what the critical number of players that need to be spammed is, so that the pop-ups aren't useless.

Quickplay mode means, that players will get exactly one MM pop-up. It will be with maximum queue width and shown to everybody. If it doesn't find a match, they are kicked out of the queue again. Nobody is removed from the CF room during this process, so if the MM fails they automatically get to play another round. Them only getting one single MM invitation will assure that whatever button they press, it doesn't lead to further pop-ups.

The only remaining disadvantage about clicking Accept I could imagine is having to wait 30 seconds while being unable to do anything. How about reducing the timer to 10 seconds?

Besides that, there will be a quick no-vote period after each autohost battle ends. This should allow some post-game discussion without degenerating into vote-spam. People stuck in the MM window won't miss out on any votes. Preferably, Chobby would also put the MM window into the background after clicking accept.

DeinFreund:
> The players are not just being trained to decline in order to avoid popups, they are also being trained to avoid missing the CF game. Players will click accept without knowing what it entails, this will lead them to abandon their MM game. This puts the other players, those who wanted to play the MM game, in the position of missing the CF game because it took them too long to figure out that their MM game was dead. This cannot be ignored and there is no helpful text box that will fix the issue.

Unless you have any idea on how to fix this issue, I agree that we'll have to disable this system if it happens. Otherwise we'd just be poisoning the MM pool for everyone. 

In that case, we could still turn this into an opt-in (poweruser) feature. Then those opt-in players would be offered MM battles when available every time their matches ends.

> The big big issue with this proposal is that I don't see what to do when the "MM accept/decline" window appears, shortly followed by the CF battle starting.

This would not be a problem, as those opt-in players would still only be offered matches once after each battle. If you want to spend a longer time in MM, I think pressing the button manually is the better solution than having lots of powerusers accidentally afking in MM.

GoogleFrog:
So your solution is to freeze the battleroom vote interaction for 30 seconds while the MM runs? Won't that just annoy people? Are you aware that I am contributing no work for anything until mid October?

PR #2418, opened 2018-09-27
only count balanced custom matches for elo

DeinFreund:


sprunk:
IMO disable elo in custom altogether

PR #2432, opened 2018-10-22
detect quitters/afk in MM battles before start, abort the battle and ban the players

DeinFreund:
I'm not too sure about the workings of the dedicated server wrapper, so please check this.

The idea is that the server checks if all players placed their commanders before the game started. In case of a matchmaker battle, not doing so will lead to a MM ban and the game being aborted. This should stop all those MM battles where somebody drops right at the start, for example due to technical issues on their first game launch.

I've added a check to make sure this only happens if the game was force started due to a timeout. If the players want, they can still decide to play an unbalanced match using `!force`.

### DIFF
--- a/Shared/LobbyClient/DedicatedServer.cs
+++ b/Shared/LobbyClient/DedicatedServer.cs
@@ -486,8 +486,8 @@ namespace LobbyClient
             try
             {
                 var timeSinceStart = DateTime.UtcNow.Subtract(Context.StartTime).TotalSeconds;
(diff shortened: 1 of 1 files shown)


sprunk:
The -20s tweak doesn't look like it belongs here, otherwise looks good

DeinFreund:
Do you not want the -20 sec tweak to happen or is it just that it doesn't fit the PR?

sprunk:
I don't mind it but it's hardly relevant to detecting people who did not place.

PR #2433, opened 2018-10-25
Send additional data to Chobby to be independent from steam overlay

DeinFreund:


ChobbyLauncher/SteamClient.cs
```diff
@@ -241,7 +241,20 @@ private string GetClientAuthTokenHex()
...
 
 
-        private List<ulong> GetFriends()
+        private List<SteamFriend> getFriends()
+        {
+            if (IsOnline)
+            {
+                return GetFriendIDs().Select(x => new SteamFriend()
+                {
+                    Name = SteamFriends.GetFriendPersonaName(x),
```
Licho1: Add avatar

DeinFreund: But avatar would have to be downloaded to an image file in the lobby?! Is it worth doing this for all your steam friends?

DeinFreund:
Added avatar, detailed friend message is described in https://github.com/ZeroK-RTS/Chobby/issues/654

GoogleFrog:
Just add a dummy function to the lobby interface if you are just being blocked by chobby not handling this.

Licho1:
@GoogleFrog  if we want this, it needs some cooperation on your side

PR #2452, opened 2018-12-11
 Block incorrect battle mode and fix ZKLobbyServer Crash

kreuwont:
Im sorry for for 4 excess commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -211,4 +211,3 @@ Zero-K.info/img/missions/169.png
 .vs
 
 Thumbs.db
-/Zero-K.plog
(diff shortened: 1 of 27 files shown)


DeinFreund:
What action lead to the Mode being invalid? I don't think the lobby should allow choosing invalid modes in the first place.

sprunk:
>I don't think the lobby should allow choosing invalid modes in the first place.

In this case the PR title says ZKLobby which is a rotting carcass at this point. Anybody can make their own lobby so garbage data has to be handled well anyway.

PR #2471, opened 2019-02-09
Fix ZeroK-RTS/Zero-K#3323

sprunk:
@DeinFreund

### DIFF
--- a/ZkLobbyServer/SpringieInterface/BattleResultHandler.cs
+++ b/ZkLobbyServer/SpringieInterface/BattleResultHandler.cs
@@ -109,7 +109,16 @@ namespace ZeroKWeb.SpringieInterface
 
             db.SaveChanges();
 
-            if (sb.ApplicableRatings == 0)
+            /* Ideally, updating the ratings would also take care of updating level:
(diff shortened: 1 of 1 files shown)


DeinFreund:
See https://github.com/ZeroK-RTS/Zero-K-Infrastructure/commit/0ea610342add7442d490b8afdeac9944481c0d5e

sprunk:
Nice, I thought the ratings were already pre-filled. On the other hand this somewhat ruins my confidence that this will actually fix that other issue.

DeinFreund:
Could you show me a replay with a draw? I wasn't aware that they're actually possible.

sprunk:
Do you want a "real, natural" replay or just a display of game mechanics?

One path to do this is to have all units die at the same time (run two identical starting commanders at each other). I don't think I've personally seen this happen in a real game but querying the DB for a PvP game where everybody died might reveal something.

Another is to have one person drop and have the other resign when presented with the opportunity to perform the inactivity win. The game interprets this as a draw and I believe I've seen this at least once but AFAIK the server misinterprets this as a victory for the dropper (because the dropper technically never died) so it might be hard to query the DB to find this.

PR #2489, opened 2019-02-18
Don't autospec on start

DeinFreund:


GoogleFrog:
Sure, why is it a PR?

PR #2504, opened 2019-03-05
automatically update game on autohosts, fix #2499

### LINKED ISSUE #2499: Coop autohosts is on the old ZK version

DeinFreund:


Licho1:
Try to avoid event handlers. There is a reason they arent used in server. They are leaky. You would need to make sure they are unsubscribed or they leak memory (all listeners will be kepty in RAM).

DeinFreund:
You're right, somehow I thought they would make for nicer code but it really just leaks now.

PR #2519, opened 2019-03-14
no elo: take from battle context

sprunk:
Instead of from extras output (unreliable, see https://github.com/ZeroK-RTS/Zero-K/issues/3223 )


sprunk:
obsoleted by #2628

PR #2536, opened 2019-04-05
only post pushes to `master` and `stable` to zkdev

sprunk:


sprunk:
Now compiles.

GoogleFrog:
Just move the crash reports, and truncate the massive PR sync messages.

Licho1:
There is/was "error" channel for errors

sprunk:
Nono, this is about pushing commits to WIP branches. Errors are untouched.

PR #2607, opened 2019-08-03
Update style.css

Fealthas:
Makes the website wider. So much wasted space with the default setting!
image

### DIFF
--- a/Zero-K.info/Styles/style.css
+++ b/Zero-K.info/Styles/style.css
@@ -51,7 +51,7 @@ pre {
 {
 	position: relative;
 	padding: 10px;
-	width: 1000px;
+	width: 80%;
 	margin-left: auto;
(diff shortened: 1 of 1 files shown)


DeinFreund:
I like the idea, but I doubt this alone is enough to make the design responsive. 

Here's how it would look on a 4k screen:
4k

Fealthas:
Hm...thats interesting. I have a 4k screen and it doesn't look anything like that. On mine, moving it to 80% just stretches the center column by like ~100px. Still looks normal, but a bit wider.

DeinFreund:
You might have some kind of dpi scaling active. Otherwise the CSS fairly
clearly indicates a width of 1000 pixels, less than a third of the width of
a 3840x2160 pixel screen.

PR #2625, opened 2019-09-17
Remove infolog link from battle detail view

Anarchid:
This has not worked since 2016. Current state of Springie does not just fail at saving the logs - it doesn't even [try](https://github.com/ZeroK-RTS/Zero-K-Infrastructure/search?q=InfologPathFormat&unscoped_q=InfologPathFormat).


Licho1:
isnt it better to fix the saving?

Anarchid:
Sure, but that's not something i can code blindly.

PR #2630, opened 2019-10-10
WIP: Redirect crash reports to a "crashreports" channel

Anarchid:
Should work but coded blindly and untested.

### DIFF
--- a/Zero-K.info/Controllers/GithubController.cs
+++ b/Zero-K.info/Controllers/GithubController.cs
@@ -40,7 +40,7 @@ namespace ZeroKWeb.Controllers
                     if (payload.action == "labeled"){
                         break;
                     }
-                    if (payload.repository.name="CrashReports"){
+                    if (payload.repository.name == "CrashReports"){
                         channel = "crashreports";
                     }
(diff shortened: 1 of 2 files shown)


Zero-K.info/Controllers/GithubController.cs
```diff
@@ -32,13 +32,17 @@ public ActionResult Hook()
...
+            string channel = "zkdev";
             Object[] values;
 
             switch (eventType) {
                 case "issues":
-                    if (payload.action == "labeled")
+                    if (payload.action == "labeled"){
                         break;
+                    }
+                    if (payload.repository.name="CrashReports"){
```
sprunk:
```suggestion
                    if (payload.repository.name=="CrashReports"){
```

sprunk: >should work

Anarchid: Oops

sprunk: story of my life

Anarchid: should work *now*!

PR #2642, opened 2019-12-03
Adding Guernsey (gg) flag

Retne:
Okay, first ever PR to a public project. Hopefully I'm along the right lines at least (I looked for contribute.md but couldn't find one - at least in this repo). Please be gentle.

Anyway, I'm trying to add a Guernsey flag to Zero-K - I am otherwise flagless when I play. That seems ironic given my country domain is used for so many gaming-related websites (.gg)
If this works I've even add the Jersey flag as a second PR.

Oh, and I first tried to create a branch thinking that was better than a fork, but I can't seem to do this. I assume that's not me-specific but a permissions thing. I'm curious why we can't branch - unless it's just to stop there being 100s of branches from Master?

Cheers for the game, folks!
R
(EDIT: Clarity)


DieGoldeneEnte:
The workflow on Github/Gitlab is to create a fork to work in and use pull requests to exchange changes between repositories, so every owner has full control over their repositories (also nessesary for security on my opinion). As you noted, this also reduces the amount of branches and nobody can (except people with extra permission) add files to a repository, which gets downloaded when cloning.

sprunk:
Hey, welcome. The PR is almost okay - there's an extra gg flag in the root folder, remove it and we can merge.

sprunk:
Additionally, this is the infra repository, which means just the website - for various reasons the flags also need to be replicated in https://github.com/ZeroK-RTS/Chobby/tree/master/LuaMenu/images/flags (for lobby) and https://github.com/ZeroK-RTS/Zero-K/tree/master/LuaUI/Images/flags (for ingame).

sprunk:
oops

Retne:
Thanks both for the comments - DieGoldenEnte for the confirmation and sprunk for your comments. I'm not sure if I managed to delete the root copy in time though - I just downloaded GitHub Desktop to do it as I couldn't see a way to do it via the GH website. Deleted it locally... I'm not sure if that then tied up to this PR.

PR #2646, opened 2020-01-20
remove lua from chobbyla

DeinFreund:


Licho1:
So what happens when you are offline and not on steam? It wont start then i think..
should be tested in offline mode first.

DeinFreund:
I don't think that was meant to be supported?

DeinFreund:
The engine referenced in modinfo.lua has been the wrong version for months. So using modinfo doesn't really help with offline play.

I originally made a PR that caches the last used engine in a local file, that should be more reliable than modinfo. #2008

GoogleFrog:
Playing offline without steam should work provided you have previously downloaded the right things. Does QueryDefaultEngine() always work?

Licho1:
No thats the thing. Thats why there is the lua fallback. QueryDefaultEngine asks the zero-k.info website.

DeinFreund:
The lua fallback asks chobby modinfo.lua, which currently says `engine          = "104.0.1-287-gf7b0fcc",
`

Licho1:
why did you merge it? Now it wont work offline at all .. if our server goes down ..

Licho1:
i dont think it should be merged, its also in use for testing scenarios when you dont want to run local site with ZK to respond to query engine

DeinFreund:
It runs offline with steam, it doesn't run offline without steam. For testing you can adjust the default engine in GlobalConst. The engine given by Chobby lua crashes the game.

Licho1:
ok, accepted by me then

GoogleFrog:
It should run offline without steam. Are you saying that it will simply crash now without steam?

DeinFreund:
No, that was never a supported configuration.

Licho1:
I am pretty sure it was at a point

Anarchid:
Just FYI, the issue that indirectly precipitated this PR (my inability to mkbundle the old minimal build of chobbyla with modern mono/mkbundle) was resolved to an mkbundle issue and workarounded as such.

PR #2651, opened 2020-03-02
this is why we don't stable on fridays

Anarchid:


ZkLobbyServer/ServerBattle.cs
```diff
@@ -917,7 +913,7 @@ protected virtual async Task OnDedicatedExited(SpringBattleContext springBattleC
             BlockPollsUntil = DateTime.UtcNow.AddSeconds(DiscussionSeconds);
 
 
-            if (IsAutohost || (!Users.ContainsKey(FounderName) || Users[FounderName].LobbyUser?.IsAway == true) && Mode != AutohostMode.None && Mode != AutohostMode.Planetwars && !IsPassworded)
+            if (Mode != AutohostMode.None && (IsAutohost || (!Users.ContainsKey(FounderName) || Users[FounderName].LobbyUser?.IsAway == true) && Mode != AutohostMode.Planetwars && !IsPassworded))
```
DeinFreund: This if is cancer. Also be aware that polls are still blocked for discussion time like this.

Anarchid: There are no custom autohosts for now, so if this blocks them, they are still blocked.

PR #2663, opened 2020-04-16
Map ladder: default supportlevel is Supported

sprunk:


sprunk:
Nevermind, unsupported maps are still taking ladder slots this way.

PR #2664, opened 2020-04-16
Don't put unsupported maps on map ladder

sprunk:


DeinFreund:
No

sprunk:
outballered by #2665

PR #2665, opened 2020-04-16
don't include map polls with unsupported maps in ratings calculation

DeinFreund:


sprunk:
Won't using `.All` cause unsupporting a map to retroactively invalidate some otherwise good polls? Maps sometimes get minor tweaks and prettiness changes and we unsupport the old version in that case. It's kinda like maps are smurfing but we don't really invalidate teamgames with a smurf in it.

sprunk:
Nvm it looks like it only affects the chunk between the first and the last update.

DeinFreund:
You are correct, I just wanted to minimize the computational expense at the
cost of ignoring some old polls.

Issue #2444, opened 2018-11-20
MM map veto design


GoogleFrog:
Players should be able to pick which MM maps they do not play on. Here are the components:
- Every player should have 6 DB keys called something like mapban1 through mapban6.
- Players can set these values by entering strings (with autocomplete, as seen here http://zero-k.info/Battles) into six text boxes in a small dedicated map ban page which is accessible via a link on the homepage.
- Every match size should have a number of map bans. Reasonable values for the current pool are 6 for 1v1, 3 for 2v2, 3 for 3v3 and 2 for 4v4.
- When a MM game is created, the map is chosen from the pool of maps created by banning the first X maps selected by both players to be banned. X varies depending on game size.

The backend should not care whether the maps banned by a player are MM maps, to avoid fancy checking or validation when the MM pool is updated. The UI should only consider MM maps for its autocomplete.

See: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/1552

Anarchid:
> Every player should have 6 DB keys called something like mapban1 through mapban6.

IMO: separate table, something like `AccountMapBans` with fields `AccountID`, `BannedMapResourceID`.

Licho1:
Yeah you dont do 'MapBan1, MapBan2, MapBan3" thats bad design. But i think this is just a description of the concept. 
Also imo its a pointless feature. It's a very niche feature for a very niche way to play the game.

GoogleFrog:
Sure, separate table.

@Licho1 it's a very important feature for the MM because many people don't like some maps. This is my experience from talking to just about anyone about updating the MM pool. I think the MM would see a noticeable increase in activity if people were able to queue with the surety that they won't be served the maps they most dislike.

GoogleFrog:
Any update?

DeinFreund:
I'd like an infra stable before doing any further changes to the MM. There is already a major rework for autohosts as well as some changes to the MM in the pipeline. For that we'll need to test and deploy:

https://github.com/ZeroK-RTS/Chobby/issues/654
https://github.com/ZeroK-RTS/Chobby/issues/650

Legomenon-gh:
I have MM running locally and intend to take a stab at implementing this since I'd very much like to make the MM map pool more configurable by users.

The last comment above by @DeinFreund implies autohosting is being reworked. Is this still current, and how much of a blocker is it? (e.g. how much of a merge conflict would it end up into in the future?)

[...]

PR #2682, opened 2020-06-17
Simple implementation of MM map veto

Legomenon-gh:
Addresses https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/2444. I have experience in other web frameworks but no previous experience in C#, ASP or this specific codebase, so please check for stupid errors or code improvements I'm not aware of.

The first commit contains all MVC architecture, and the second commit is only the changes to the matchmaker itself. The second commit is the one most likely to cause problems, so it can be reverted on its own.

Testing I've done locally:
- Verify multiple players can store their bans
- Verify bad input doesn't cause any problems
- Verify map bans are applied in a 1v1 scenario
- Verify setting map bans to 0 for 1v1 results in the same behavior as before my changes
- Verify bans for previous matchmaking maps are ignored
- Verify accidentally banning all the maps doesn't break matchmaking
- Verify map autocomplete still works in Battle list and elsewhere

I cannot verify actual hosting of a game locally due to insufficient memory on my machine, or more than 1v1. I don't believe there's a working test instance for matchmaking at present, so it's probably worth having someone else sanity check it locally if they can.

Room for future improvement: have separate bans for 1v1 and team games and a less spartan UI. Right now applying bans to maps only used in team games requires "losing" bans for 1v1, and selecting maps is a little confusing since all matchmaking games are returned and it's not clear which ones are only used in 1v1. This is probably fine for now since team game MM is pretty niche, and it can be worked around (by increasing the total number of bans per user or just having users editing their bans to be more targeted at team games).


Zero-K.info/Controllers/AutocompleteController.cs
```diff
@@ -108,11 +113,17 @@ public ActionResult Clans(string term)
...
-                db.Resources.Where(x => x.InternalName.Contains(term) && x.TypeID == ResourceType.Map)
+            var maps = db.Resources.AsQueryable();
+            if (mapSupportLevel != null)
+            {
+                maps = maps.Where(x => x.MapSupportLevel == mapSupportLevel);
+            }
+
+            return maps.Where(x => x.InternalName.Contains(term) && x.TypeID == ResourceType.Map)
                     .OrderByDescending(x => x.MapSupportLevel)
+                    .ThenBy(x => x.InternalName)
```
Legomenon-gh, line 126: I think applying a secondary alphabetical sort is a fine change to make in all cases, but I can restrict it to the `mapSupportLevel` filter case if necessary

DeinFreund:
I don't have time to review it in detail now, if you want I can merge it so you see how it behaves on the test server.

Legomenon-gh:
Thanks! Pushing to the test server would be great, feel free to do that whenever convenient and I'll test the bans more properly there.

Zero-K.info/Views/MapBans/MapBansIndex.cshtml
```diff
@@ -0,0 +1,26 @@
...
+    For team games, your first few unique bans will be used depending on the size of the game. </p> 
+
+<form action="@Url.Action("Update")" method="post">
+    @for (int i = 0; i < Model.maxBans; i++)
+    {
+        String mapName = i < Model.accountBans.Count ? Model.accountBans[i].Resource.InternalName : null;
+
+        <div style="margin-bottom:10px">
+            <span style="padding-right:10px;">Ban @(i+1):</span>
+            <span>@Html.TextBox("mapName", mapName, new { data_autocomplete = Url.Action("MatchMakerMaps", "Autocomplete")})</span>
```
DeinFreund, line 22: This may be a bit cleaner using the Model and `TextBoxFor`. For an example look at https://stackoverflow.com/questions/38205799/using-html-textboxfor-to-loop-through-items-in-a-model

Shared/PlasmaShared/MapBanConfig.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Collections.Generic;
+
+namespace ZkData
+{
+    public class MapBanConfig
+    {
+        private static Dictionary<int, int> BansPerPlayerByGameSize = new Dictionary<int, int>() {
```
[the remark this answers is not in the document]
Legomenon-gh: That's an excellent question! I didn't think of using the map pool size dynamically but that's a much nicer solution. The current code technically works because the size of the pool is pretty static and the team pool is bigger than the 1v1 pool, but it's more complicated, more brittle and pretty ugly, so yeah I'll fix it :D

New logic will be "take the number of maps in the pool and find the nearest integer that bans no more than a configurable 75% or 80% of the maps. Configure the number of bans displayed in the UI separately".

PR #2683, opened 2020-06-18
Improvements to matchmaking map ban

Legomenon-gh:
See https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/2682 for context


Zero-K.info/Controllers/MapsController.cs
```diff
@@ -81,10 +81,13 @@ public class MapsController: Controller
                                 out ret);
 
             if (!offset.HasValue) {
+                // Allow to open maps page with the matchmaking option already set so it can be used in links
```
Legomenon-gh, line 84: This isn't strictly necessary but allows for a nice link from the map ban page, I can take it out if it will cause issues

Shared/PlasmaShared/GlobalConst.cs
```diff
@@ -150,6 +150,9 @@ public static BasicHttpBinding CreateBasicHttpBinding()
         public const float LadderEloMinChange = 1;
         public const float LadderEloSmoothingFactor = 0.25f; //1 for change as fast as whr, 0 for no change
 
+        public const int MapBansPerPlayer = 6; // Allow users to enter this many bans in UI
```
Legomenon-gh, line 153: I'll change those numbers if anybody has strong opinions on how much of the pool should be bannable. (The number of bans shown in the UI could be increased past 6 but that would only have any impact if there are lots of shared bans across all players. 6 is the maximum number that guarantees all bans will be applied in a 1v1 game)

Zero-K.info/Views/Shared/TopMenu.cshtml
```diff
@@ -14,8 +14,7 @@
                         <li><a href="@Url.Action("Index", "Home")">Home</a></li>
                         <li><a href="@Url.Action("UnlockList", "My")">Unlocks</a></li>
                         <li><a href="@Url.Action("Commanders", "My")">Commanders</a></li>
-                        <!-- TODO: Uncomment ban link after sufficient MM testing on live server -->
-                        <!-- <a href="@Url.Action("Index", "MapBans")">Map Bans</a></li> -->
+                        <li><a href="@Url.Action("Index", "MapBans")">Map Bans</a></li>
```
Legomenon-gh, line 17: Feel free to comment this out again if it makes deploying easier

Zero-K.info/Views/MapBans/MapBansIndex.cshtml
```diff
@@ -1,25 +1,28 @@
-﻿@using ZeroKWeb.Controllers
-@model MapBansViewModel
+﻿@using ZeroKWeb.Controllers @using ZkData
```
sprunk, line 1:
```suggestion
﻿@using ZeroKWeb.Controllers
﻿@using ZkData
```

Legomenon-gh: Even better, I don't actually use the first import anymore! (I committed your change before realizing this and force pushed it away to reduce commit noise)

Zero-K.info/Controllers/MapBansController.cs
```diff
@@ -16,18 +16,25 @@ public class MapBansController : Controller
         public ActionResult Index()
         {
             var db = new ZkDataContext();
```
DeinFreund, line 18: Since you're already using ToList() it seems like you want to avoid db queries during website rendering. To assure this, you could use `using (var db = new ZkDataContext()) {...}`. I generally prefer it since it avoids hidden db queries.

ZkLobbyServer/MatchMaker/MatchMaker.MapVetoer.cs
```diff
@@ -35,9 +35,15 @@ public string SelectMap(List<string> candidates)
...
             private List<string> CandidatesAfterBans(List<string> candidates, IEnumerable<IGrouping<int, AccountMapBan>> bans) {
                 var bannedMaps = new HashSet<string>();
-                int bansPerPlayer = MapBanConfig.GetPlayerBanCount(Battle.Size);
 
-                // For each player in the battle, go through their map bans and add the first X bans not already used by another player.
+                // Ban up to X% of all maps in the pool, and distribute those bans evenly across all players, rounding down.
+                // This could be static configuration on the queue itself but the map ban logic may take into account the actual
+                // ban distribution of each player in the future and this keeps all the banning logic outside the main MatchMaker class.
+                var maximumBans = GlobalConst.MaximumPercentageOfBannedMaps * candidates.Count;
+                var bansPerPlayer = Math.Truncate(maximumBans / Battle.Size);
```
DeinFreund, line 43: Truncate is an interesting choice, why not floor?

Legomenon-gh: No reason whatsoever :D Good catch

Zero-K.info/Controllers/MapBansController.cs
Legomenon-gh: Is there an existing system to log all ZkDataContext DB queries by the way? (I didn't see anything built in but maybe it's done via magical configuration. I've done it via the Management Studio tool but MS SQL is surprisingly poor at logging SQL queries)

DeinFreund: I don't know of anything. If necessary I end up using the Microsoft SQL Server Profiler.

Legomenon-gh:
Sanity checked again on the test server

PR #2687, opened 2020-07-02
Hide Post comment button when making/editing post

Legomenon-gh:
Is showing the "Post comment" control both above and below the content intentional? I assume it is

### DIFF
--- a/Zero-K.info/Controllers/ForumController.cs
+++ b/Zero-K.info/Controllers/ForumController.cs
@@ -171,7 +171,8 @@ namespace ZeroKWeb.Controllers
 
             if (!model.DisablePostComment)
             {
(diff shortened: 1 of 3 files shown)


Zero-K.info/Controllers/ForumController.cs
```diff
@@ -166,9 +166,14 @@ public ActionResult DeletePostPrompt(int? postID)
                 if (filterAccountID.HasValue) posts = posts.Where(x => x.AuthorAccountID == filterAccountID);
             }
 
-            model.Data = posts.OrderBy(x=>x.ForumPostID);
+            model.Data = posts.OrderBy(x => x.ForumPostID);
             model.Thread = thread;
 
+            if (!model.DisablePostComment)
+            {
+                model.DisablePostComment = thread.IsLocked || thread.ForumCategory.ForumMode == ForumMode.Maps || thread.ForumCategory.ForumMode == ForumMode.Missions || thread.ForumCategory.ForumMode == ForumMode.SpringBattles|| thread.ForumCategory.ForumMode == ForumMode.Clans || thread.ForumCategory.ForumMode == ForumMode.Planets;
```
Legomenon-gh, line 174: I can clean this up further but wanted it out of the view at least (unfortunately this isn't the same as "IsNormalThread")

Legomenon-gh:
If you have an idea on how to infer that, sure. I don't see a way obviously or I'd have done that already

PR #2688, opened 2020-07-03
Display map size in map polls

### LINKED ISSUE #2686: Would be good information to show the sizes of the maps in !map vote
See title.

Legomenon-gh:
Fixes #2686

### DIFF
--- a/ZkLobbyServer/ServerBattle.cs
+++ b/ZkLobbyServer/ServerBattle.cs
@@ -963,7 +963,8 @@ namespace ZkLobbyServer
                     pickedMaps.Add(map.ResourceID);
                     options.Add(new PollOption()
                     {
-                        Name = map.MapNameWithDimensions(),
+                        Name = map.InternalName,
+                        DisplayName = map.MapNameWithDimensions(),
                         URL = $"{GlobalConst.BaseSiteUrl}/Maps/Detail/{map.ResourceID}",
                         ResourceID = map.ResourceID,
                         Action = async () =>
--- a/ZkLobbyServer/ServerBattle.cs
+++ b/ZkLobbyServer/ServerBattle.cs
@@ -963,7 +963,7 @@ namespace ZkLobbyServer
                     pickedMaps.Add(map.ResourceID);
                     options.Add(new PollOption()
                     {
-                        Name = map.InternalName,
+                        Name = map.MapNameWithDimensions(),
                         URL = $"{GlobalConst.BaseSiteUrl}/Maps/Detail/{map.ResourceID}",
                         ResourceID = map.ResourceID,
                         Action = async () =>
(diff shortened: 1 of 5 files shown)


ZkLobbyServer/autohost/Commands/CmdMap.cs
```diff
@@ -53,7 +53,6 @@ public override string Arm(ServerBattle battle, Say e, string arguments = null)
                 alternativeMap = MapPicker.FindResources(ResourceType.Map, arguments, MapSupportLevel.Supported, true).FirstOrDefault();
             }
 
-
             if (Map.MapSupportLevel >= MapSupportLevel.Supported)
             {
                 return $"Change map to {Map.InternalName}?";
```
sprunk, line 58: what about here?

ZkLobbyServer/ServerBattle.cs
```diff
@@ -963,7 +963,7 @@ private async Task<bool> CreateMultiMapPoll()
                     pickedMaps.Add(map.ResourceID);
                     options.Add(new PollOption()
                     {
-                        Name = map.InternalName,
+                        Name = map.MapNameWithDimensions(),
```
sprunk, line 966: Check how this interacts with Chobby. I believe it might be using the raw name to get minimaps so there's a worry that adding some metadata like size would ruin that (see https://github.com/Spring-Chobby/Chobby/blob/95f35a600aca3a0b2c338932fda133086a8c86c0/LuaMenu/widgets/gui_battle_room_window.lua#L1317 )

Legomenon-gh: Yeah I'm pretty sure you're right :( Minimaps do not work in my local env even without any changes so I didn't catch that, I'll close this PR for now

sprunk: Does the single map poll show minimap? IIRC it didn't in which case it should be fine to add that.

Legomenon-gh: The multi map pool (i.e. `!pool map) does use minimaps so that's a problem, and not using the correct name is too likely to result in breakage (I did not realize the internalName was used as a key too in Chobby)

sprunk: Perhaps the size should just be a separate key transferred along the other options? That way Chobby can draw it somewhere.

```c#
options.Add(new PollOption()
{
    Name = map.InternalName,
    MapWidth = map.MapWidth,
    MapHeight = map.MapHeight,
```

Legomenon-gh: Yeah I'll probably do that (or a `tooltip` property instead) after stable is cut

Legomenon-gh:
I've verified this with https://github.com/ZeroK-RTS/Chobby/compare/master...Legomenon-gh:poll-tooltip?expand=1. I'll open up the Chobby PR if this side looks good (the two can be deployed separately)

PR #2717, opened 2020-08-16
Improvements to modaction interface

Aquanim:
- PMs sent to modactioned users explicitly state that a modaction without any bans/mutes/etc is a warning, and otherwise explicitly lists the penalties and duration
- User page on site explicitly states that a modaction without any bans/mutes/etc is a warning
- Vote ban is correctly displayed on user page
- Several modaction types are correctly printed in zkadmin channel
- Added new checkbox and database column for sending messages via the modaction interface that are not intended to be penalties; instead of the big PENALTY message this puts a link on the user page to chatting with Nightwatch, which as far as I know makes it private? Even if not this still seems useful


Aquanim:
Note that the message-only option is (or at least should be) overridden by any other penalty being applied.

PR #2723, opened 2020-08-28
"Blind" attempt at fixing NW report droppage

sprunk:
Sometimes messages can get through which suggests a race condition, nobody else has any better idea so won't hurt to try.


Licho1:
There was probably no await on purpose, to make it async from the point of user

sprunk:
8d4a86937e53ac89454016de2aa37ee34c03a0c9 my vision is augmented

Licho1:
The referenced commits I am not sure since it will require async controller too..

sprunk:
@DeinFreund ^

PR #2760, opened 2020-11-09
Tell Spring games are ZKLS-hosted

sprunk:
My use-case is that this lets the cheat-menu widget know it has to use `!hostsay /foo` instead of `/foo`.
Another one would be avoiding sending data via `SPRINGIE:FOO` if we know there is no lobbyserver to receive it (awards, modstats, etc).


Licho1:
Looks ok, but consider better name or at least casing in the name.
Also isn't there already a way to detect it?
Also consider having each thing separated. Like
 
"cheatPrefix"   = "!hostsay"
and
"relayChat" 

That would give us more control and flexibility in the future.

Licho1 (approved):

PR #2764, opened 2020-11-14
Added checkbox so users can tag uploaded maps as special.

Aquanim:
I can't easily test this locally so somebody who actually understands infra should check this before pulling.

### DIFF
--- a/Zero-K.info/Controllers/MapsController.cs
+++ b/Zero-K.info/Controllers/MapsController.cs
@@ -426,7 +426,7 @@ namespace ZeroKWeb.Controllers
         }
 
         [Auth]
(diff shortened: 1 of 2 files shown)


DeinFreund:
This will likely conflict with @maackey's changes.

Aquanim:
For reference the reason for this PR is that retrospectively tagging a bunch of ZeroWars replays as noelo is a pain in the neck.

DeinFreund:
Yes I totally understand. That's exactly the use case I'd expect it to be used for. Just asking maackey to check whether he wants to merge his changes soon, and in this case would rather just include this in his UI overhaul.

maackey:
This is only a few lines it looks like it will be pretty easy to merge

PR #2776, opened 2020-12-01
Automatically notify admin channel of kicks in autohosts

Aquanim:


ZkLobbyServer/autohost/Commands/CmdKick.cs
```diff
@@ -44,13 +44,75 @@ public override string Arm(ServerBattle battle, Say e, string arguments = null)
                 battle.Respond(e, "Only players can invoke this during a game");
                 return null;
             }
+
+            if (battle.IsAutohost)
```
Licho1, line 48: It would be nicer to encapsulate it to another function like "ReportToAdmins" because it has nothing to do with actual poll logic.

ZkLobbyServer/autohost/Commands/CmdKick.cs
```diff
@@ -44,13 +44,75 @@ public override string Arm(ServerBattle battle, Say e, string arguments = null)
...
+
             return $"Do you want to kick {target}?";
         }
 
 
         public override async Task ExecuteArmed(ServerBattle battle, Say e)
         {
-            if (battle.spring.IsRunning) battle.spring.Kick(target);
+
+            if (battle.IsAutohost)
```
Licho1, line 84: It would be nicer to encapsulate it to another function like "ReportToAdmins" because it has nothing to do with actual poll logic.

Aquanim:
I can make those changes. Probably tomorrow though.

Aquanim:
Should be done now.

Licho1 (approved):

PR #2779, opened 2020-12-09
Add server management of game modes

Licho1:
This adds server-side management of custom game modes.

List and management is super simple and looks like this:

image

Detail of custom mode looks like this:
image

And editing/creation:
image

Steam distribution saves all featured game modes to CustomModes folder.
Non steam chobbyla downloads all featured game modes to CustomModes folder on startup.

For "play this mode" command is sent to lobby:

```chat/battle@select_customGameMode:shortName```

New lobby messages are added:
```
    [Message(Origin.Client)]
    public class GetCustomGameMode
    {
        public string ShortName;
    }
```
And response
```
    [Message(Origin.Server)]
    public class CustomGameModeResponse
    {
        public string ShortName;
        public string DisplayName;
        public string GameModeJson;
    }
```

## Open questions
 
 * do we want to store extra description to show on website with more detailed mode info (with forum formatting? or extract form json?)
 * is the site to lobby command ok? 
 * the name "custommode" vs "game mode" etc.. perhaps we should just call them "Mods" ?


Licho1:
@GoogleFrog  your opinion for open questions is needed

GoogleFrog:
People can, and probably should, describe the custom mode in the first post of its forum thread. Also, I don't see any reason to hide the contents of the json from anyone. Perhaps add a 'view details' button that expands to show the full json of the custom mode.

Just to check, is the command for "play this mode" a SiteToLobbyCommand, similar to what is already handled here: https://github.com/ZeroK-RTS/Chobby/blob/master/libs/liblobby/lobby/interface_zerok.lua#L1759 ? If it is then it looks ok, although it seems a bit convoluted. Is the intended workflow as follows:
 - User click "play this mode"
 - Site sends lobby a SiteToLobbyCommand with the shortname.
 - Lobby sends server a GetCustomGameMode with the shortname.
 - Server sends lobby a CustomGameModeResponse with the json. It saves this json locally for future access in the user's local custom mode list, and uses the json to set a custom mode (potentially opening a battle to do so).

I struggle a bit with naming them. Technically a 'custom mode' could be a map, mutator, set of modoptions, or even a new game. The word 'mod' in Spring usually means something like mutator/game, so it could be confusing for people who look into the Spring documentation. I suppose we could just call them "Mods" and ensure that we say mutator/game when referring to technical mutators/games. It would be nice to have a single word like "Mod" and players from elsewhere would have a better idea of what "Mod" means.

Licho1:
Problem with forum post is that posts are sorted newest on top. So you will see some non-relevant comments instead of opening post. I guess it could be reversed in this case but then it's hard to check "whats new".

Yeah the click sends SiteToLobbyCommand. I did it this way mainly for consistency. If you think its better we could instead send directly CustomGameModeResponse (call it "HostCustomGameMode") along with json data .

We could show json data but they should be click hidden, because its highly technical and can be huge. 
Perhaps there could be "download manually" button instead?

Regarding name: yes it's a problem and I am certain "Mods" is used in many places in the code already. But we could use different internal name and public name. 

Btw all of this could also be achieved via steam workshop without any interference from us. 
People could just upload their "mod" json file as a workshop item. 

I am thinking that we should be encouraging steam workshop more. It has better social system and its probably good for longevity and visibility.

GoogleFrog:
Could you add a description field that essentially makes a forum post that is pinned above the "submit comment" field?

I want to have a button that shows to json to make it easier for people to see what goes into defining these mods.

I don't know how to encourage the steam workshop.

maackey:
Description should be in the body of the page -- not part of the forum. 
I also think it would be beneficial to show the full json on the page. 
"Mods" is a good public name that people are familiar with.
I'm not familiar with the workshop/steam integration so much... is it possible to include a link to the workshop?
I'm also not familiar with the lobby commands so I don't have any comment on that.

I expect a lot of visual changes once I get a hold of it since this is pretty similar to the replays page. For reference, here's where I currently am with the replays page. (lots of things need tweaking, and "play now" button in top right isn't there because I don't actually have any battles to work with)
image

PR #2781, opened 2020-12-19
Allow users who log in through Steam to set a password.

Aquanim:
Revisit this when Steam is no longer having issues.


Aquanim:
Possibly there should be more capability for users to link or unlink their accounts from Steam themselves to deal with future problems of this kind.

Aquanim:
Looks like this is being solved more generally by #2783 .

PR #2787, opened 2021-01-26
Allow searching replays by is-MM and rating type.

Aquanim:
Also displays (primary) rating type on each replay page to non-admins.


Shared/PlasmaShared/ISpringieService/RatingCategory.cs
```diff
@@ -11,4 +11,8 @@ public enum RatingCategoryFlags
     {
         Casual = 1, MatchMaking = 2, Planetwars = 4
     }
+    public enum RatingSearchOption
+    {
+        Any = 0, Casual = 1, Competitive = 2, Planetwars = 3, None = 4
```
maackey, line 16: minor - I'd group up "None" and "Any" just in case additional options are added in the future, it'll be less fragmented

Zero-K.info/Controllers/BattlesController.cs
```diff
@@ -138,6 +140,34 @@ public enum AgeOption
...
+                        break;
+                    case RatingSearchOption.None:
+                        q = q.Where(b => b.ApplicableRatings == 0);
+                        break;
+                    default:
+                        // The default case being no filtering should be safe enough.
+                        break;
+                }
+            }
+
```
maackey, line 170: I'm not sure why a temporary variable needs to be created.
`q = q.Where(b => b.IsMatchMaker == model.Matchmaker`
Other places use nullable booleans or int values to represent similar concepts -- I personally think YesNoAny type is dumb, especially if it doesn't simplify working with nullable booleans, but at least its consistent with the other stuff on this page. 

The conditional is redundant around the switch statement. 
`switch (model.Rating)
{
...
default: 
// default case has no filtering and is effectively equivalent to "any"
}`

Zero-K.info/Views/Battles/BattleDetail.cshtml
```diff
@@ -92,9 +92,37 @@
...
+                        break;
+                }
+            }
+            else
+            {
+                <text>None</text>
+            }
+        }
+    </td>
+</tr>
```
maackey, line 122: Shouldn't Model.GetRatingCategory() return "None" if Model.IsRatedMatch() is false? That would greatly simplify the view, which IMO shouldn't have lots of logic embedded -- it gets messy and hard to style.

maackey (commented):
Just a few minor suggestions after glancing it over.

Zero-K.info/Controllers/BattlesController.cs
```diff
@@ -138,6 +140,34 @@ public enum AgeOption
...
+            if (model.Rating != RatingSearchOption.Any)
+            {
+                switch (model.Rating)
+                {
+                    case RatingSearchOption.Competitive:
+                        q = q.Where(b => b.ApplicableRatings.HasFlag(RatingCategoryFlags.MatchMaking));
+                        break;
+                    case RatingSearchOption.Casual:
+                        q = q.Where(b => b.ApplicableRatings.HasFlag(RatingCategoryFlags.Casual));
```
Licho1, line 157: Has flag wont have translation to SQL so this wont work. It would crash unless you dod ToList before. And if you do you are copying battles to memory

Licho1 (changes requested):

Aquanim:
> minor - I'd group up "None" and "Any" just in case additional options are added in the future, it'll be less fragmented
Good point.

> Shouldn't Model.GetRatingCategory() return "None" if Model.IsRatedMatch() is false? That would greatly simplify the view, which IMO shouldn't have lots of logic embedded -- it gets messy and hard to style.
That seems like a reasonable take to me, but outside the scope of this PR.

> The conditional is redundant around the switch statement.
It is, but that applies to the conditionals around some of the other switch statements previously existing in this file as well so I followed the convention. I assume the purpose is to make the categories easier to extend in the future without making mistakes.

> Has flag wont have translation to SQL so this wont work. It would crash unless you dod ToList before. And if you do you are copying battles to memory
https://github.com/dotnet/efcore/issues/2852 suggests that HasFlag does have a translation as of 2016. Certainly it did not throw an error like b.IsRatedMatch() did when I used it in the same place. Can you confirm that this is a problem?

Zero-K.info/Controllers/BattlesController.cs
Licho1: Nullable value is fine, we just lacked gui component to represent that ..

Zero-K.info/Controllers/BattlesController.cs
Licho1: To resolve this use condition like this (b.ApplicableRating & RatingCategoryFlags.Casual) = RatingCategoryFlags.Casual

Licho1 (approved):

PR #2794, opened 2021-02-08
Improve support on distros using 256 color terminals

Porkch0p:
Most Linux installs are now using 256 color terminals. The launcher conflicts with this. Disables 256 color support at launch of script to fix or improve [Chobby#755](https://github.com/ZeroK-RTS/Chobby/issues/755#issuecomment-772746566) (may need to be ported to itch separately)


Porkch0p:
The choice of "xterm" in the script shouldn't cause any problems, since even if it weren't available, an invalid terminal also falls back correctly in every instance I've seen.

PR #2799, opened 2021-02-28
Require one extra vote for resign polls to pass

Legomenon-gh:
This change makes resign require unanimous approval for small team games (up to 4v4) and makes resign a little harder for large team games, but can be tweaked depending on the exact desired behavior.

Based on feedback in https://zero-k.info/Battles/Detail/1060977

I've sanity checked this to work for 1 and 2 players but haven't fully tested past that, I will do that once I have sign off (it should work fine, but arithmetic errors are embarrassing and testing with 4 accounts locally is a pain)

### DIFF
--- a/ZkLobbyServer/ServerBattle.cs
+++ b/ZkLobbyServer/ServerBattle.cs
@@ -604,7 +604,10 @@ namespace ZkLobbyServer
                 url = $"{GlobalConst.BaseSiteUrl}/Maps/Detail/{(unwrappedCmd as CmdMap).Map.ResourceID}";
                 map = (unwrappedCmd as CmdMap).Map.InternalName;
             }
-            poll = poll ?? new CommandPoll(this, true, true, unwrappedCmd is CmdMap, map, unwrappedCmd is CmdStart);
+
+            var voteMargin = unwrappedCmd is CmdResign ? (unwrappedCmd as CmdResign).RequiredWinMargin : 1;
+
+            poll = poll ?? new CommandPoll(this, true, true, unwrappedCmd is CmdMap, map, unwrappedCmd is CmdStart, voteMargin);
             options.Add(new PollOption()
             {
                 Name = "Yes",
--- a/ZkLobbyServer/autohost/Commands/CmdResign.cs
+++ b/ZkLobbyServer/autohost/Commands/CmdResign.cs
@@ -17,6 +17,10 @@ namespace ZkLobbyServer
         public override string Shortcut => "resign";
         public override AccessType Access => AccessType.IngameVote;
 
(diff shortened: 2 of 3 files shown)


ZkLobbyServer/ServerBattle.cs
```diff
@@ -604,7 +604,10 @@ public async Task<bool> StartVote(BattleCommand cmd, Say e, string args, int tim
                 url = $"{GlobalConst.BaseSiteUrl}/Maps/Detail/{(unwrappedCmd as CmdMap).Map.ResourceID}";
                 map = (unwrappedCmd as CmdMap).Map.InternalName;
             }
-            poll = poll ?? new CommandPoll(this, true, true, unwrappedCmd is CmdMap, map, unwrappedCmd is CmdStart);
+
+            var voteMargin = unwrappedCmd is CmdResign ? (unwrappedCmd as CmdResign).RequiredWinMargin : 1;
```
Legomenon-gh, line 608: There's a couple other ways to do this, I went with the simple cast (as was done with the map command seen above) since only resign votes should require a different vote threshold

ZkLobbyServer/autohost/Commands/CmdResign.cs
```diff
@@ -17,6 +17,10 @@ public class CmdResign : BattleCommand
         public override string Shortcut => "resign";
         public override AccessType Access => AccessType.IngameVote;
 
+        // We might want different logic to determine the success of a resign vote based on the number of players.
```
Legomenon-gh, line 20: This can turn into a method to apply different thresholds for different game sizes, or be something like 66% instead, whichever GoogleFrog wants

GoogleFrog:
As I said on the post, I like the idea.

Legomenon-gh:
Yeah I wasn't sure if you liked the idea for small teams alone or all games, but at least it should be simple to tweak moving forward.

This has been tested more properly now

Licho1:
I think that for big games, we should instead ask only people who are not resigned already. Often people who already resigned vote to resign the rest so that they can have another game.

GoogleFrog:
> I think that for big games, we should instead ask only people who are not resigned already. Often people who already resigned vote to resign the rest so that they can have another game.

Agreed. People would complain initially about having a few players left extending the game, but in that case the resigners should either not resign or accept that people are happy with playing on and seeing the explosions.

Legomenon-gh:
I'm not sure if this is general talk or feedback on this PR, but I am not changing the large team game behavior since I cannot test it locally and this is separate from the "player can be forcefully resigned out of a 3v3/4v4" problem

GoogleFrog:
I think the scope of this PR is good.

PR #2801, opened 2021-03-08
Add sanity check on map ban insertion

Legomenon-gh:
The check on the maximum map ban length on update was dropped after a refactor, this commit reapplies it. The map banner will
not apply bans past a maximum threshold (currently 75%) so inserting excess bans will not break matchmaker but would give the player a couple of extra bans.

### DIFF
--- a/Zero-K.info/Controllers/MapBansController.cs
+++ b/Zero-K.info/Controllers/MapBansController.cs
@@ -58,6 +58,11 @@ namespace ZeroKWeb.Controllers
                 return Content("The same map cannot be banned multiple times.");
             }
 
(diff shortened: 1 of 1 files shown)


Legomenon-gh:
Just noticed this bug, apologies. A quick check on the database would be good to trim any existing extra map bans (I doubt anybody would bother making custom POST requests on purpose to do this, but there could always be a bug somewhere too)

sprunk:
@Anarchid or @Aquanim ping for SQL

DeinFreund:
image

I suppose this means nothing has to be done?

sprunk:
Seems so, thanks

Issue #2780, opened 2020-12-10
Waitlist players count towards vote requirements


Arch-Shaman:
Replication:
1.) Make 1v1 room
2.) put 4 players into that room
3.) Votes now require 3 people to start

PR #2815, opened 2021-04-16
Clamp voter count to the room maxPlayers

Legomenon-gh:
This is a bandaid fix for waiting list players being included in the quorum for polls. A full fix would require also preventing waiting list players from starting or participating in polls.

Partly addresses https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/2780

### DIFF
--- a/ZkLobbyServer/ServerBattle.cs
+++ b/ZkLobbyServer/ServerBattle.cs
@@ -607,7 +607,12 @@ namespace ZkLobbyServer
                 map = (unwrappedCmd as CmdMap).Map.InternalName;
             }
 
-            var numVoters = Users.Values.Count(x => selector(x.Name) == null);
+            // Players in the waiting list are in a limbo between spectating and not. The lobby is responsible
+            // for determining who gets to play and it is not trivial to have the server know who should be
+            // allowed to participate in polls when a wait list is involved in sync with the lobby.
+            // For now, ensure that a poll does not require more voters than there are active players.
+            // TODO: Prevent waiting list players from starting or participating in polls
+            var numVoters = Math.Min(MaxPlayers, Users.Values.Count(x => selector(x.Name) == null));
             var voteMargin = unwrappedCmd.GetPollWinMargin(this, numVoters);
 
             poll = poll ?? new CommandPoll(this, true, true, unwrappedCmd is CmdMap, map, unwrappedCmd is CmdStart, voteMargin);


Legomenon-gh:
I'd like to know if I'm correct on the behavior of the waiting list before I dig too much deeper, or if I missed something and it's easier than I think it is.

If my understanding is correct, is it worth having a full fix, and do you have any pointers to the safest/easiest way to deal with the waiting list in polls?

sprunk:
> I'd like to know if I'm correct on the behavior of the waiting list

The comment on your PR says "The lobby is responsible for determining who gets to play" but actually the server does: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/26698dadedc96bc2515ad15a00bef9b26c24cbb8/ZkLobbyServer/ServerBattle.cs#L535-L538

It's still not obvious on who should vote because the queue order is set once when you join the room and doesn't reset when you spec/unspec. Consider this situation:
 
 * room with 2 slots.
 * player A joins (gets queue order 1), sets Playing.
 * player B joins (gets queue order 2), sets Spectating.
 * player C joins (gets queue order 3), sets Playing.
 * at this point A and C are set to be playing and B is a spec. If a game is started, A and C will play.
 * then player B changes from Spectating to Playing (unspecs). This pushes C back into the waiting list (because 2 < 3)! If a game is started, A and B will play.

Legomenon-gh:
Indeed, this part of the comment is misleading, I've updated it and clarified that this does not affect ingame polls. Feel free to reword it further for clarity too.

(I don't think it would be too arduous to filter out voters by whether they will get to play in the current state of affairs, but it's quite a bit of extra code that shouldn't be hit very often, so curious if it's worth doing)

sprunk:
Feel free to, it would prevent some cases that are clearly bad (for example in a 1v1 you shouldn't be able to decide the map / modoptions and then start just because you brought a homie). This PR is already a standalone improvement though so no hurry.

Legomenon-gh:
Yeah I'm probably not filtering out wait list players from polls until/unless there's examples of users gaming that system, since it's relatively painful to test locally and ensure that the fix is very low risk (i.e. I don't want to break the teams all welcome room in a way I couldn't replicate in dev)

sprunk:
Given that you can vote "as a spec" anyway (unspec, cast a vote, spec back) and people are using that power fairly responsibly it's probably fine to let the waiting list vote for now.

sprunk (approved):

PR #2816, opened 2021-04-22
(Check builds first) Stable for resign thresholds

sprunk:
For the resign threshold changes, should probably be merged during the night between Thursday and Friday so as to have it available for the weekend. Remember to check master builds first though because apparently they don't build.

### DIFF
--- a/ZkLobbyServer/autohost/Commands/CmdResign.cs
+++ b/ZkLobbyServer/autohost/Commands/CmdResign.cs
@@ -17,10 +17,6 @@ namespace ZkLobbyServer
         public override string Shortcut => "resign";
         public override AccessType Access => AccessType.IngameVote;
 
-        // We might want different logic to determine the success of a resign vote based on the number of players.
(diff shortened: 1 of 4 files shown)


Aquanim:
Master builds, pulling to stable.

ZkLobbyServer/autohost/Commands/CmdResign.cs
```diff
@@ -70,5 +66,17 @@ public override RunPermission GetRunPermissions(ServerBattle battle, string user
             }
             return RunPermission.None;
         }
+
+        public override int GetPollWinMargin(ServerBattle battle, int numVoters)
+        {
+            // Require unanimous vote for resigning up to 3v3 inclusive
+            if (numVoters > 3)
```
Legomenon-gh, line 73: I wonder if it'd be beneficial to return just 0 instead of 1 starting at some large team sizes since simple majority is a sufficient bar to clear (doubt it's worth doing now with gameside resign coming though)

sprunk: Doesn't 0 mean that you'd only need N votes out of 2N eligible?

PR #2846, opened 2021-10-06
less obsolete unitguide link

Anarchid:


sprunk:
`manual.zero-k.info` is used to host some resources (images on other wikipages etc) IIRC.

Issue #2848, opened 2021-10-11
Very Simple League Idea


GoogleFrog:
Here is a league idea with maximal code reuses. It involves:
- Add/copy/rename the planetwars ladder.
- Remove all games from this ladder.
- Add a system that can be enabled in dynamic config that autotags 1v1 MM games for this ladder.
- Add some sort of additional visibility to this ladder so people feel like its the league (maybe add/replace the front page box with it?).
- Announce the league and the system above.
- Flip the dynamic config switch to start recording games.
- Wait the number of months decided upon at the start of the league.
- Save the state of the ladder in some sort of static format (screenshots, copying text to a wiki page, etc).
- Click the button (ideally) or do the DB query (less ideal) that clears all games from this ladder.
- Announce a new league, flip the switch, etc.

PR #2850, opened 2021-10-11
WIP: seasonal ladder rating

sprunk:
Don't merge! Lacks migration.

To do:
> We probably need a lobby UI matchmaker tickbox that toggles whether the player is queuing for a league game or not. The game would only count if both players say it is a league game. I don't know whether this would break the league, but it seems wise to have the capability.


GoogleFrog:
Remind me, what does this implement and what would the lobby need?

sprunk:
This implements a fourth rating, "Ladder" (in addition to MM, Casual aka custom, and PW).
There is a new dynamic config entry on the website, which is a boolean toggle on whether a league is running.
If the bool is toggled, 1v1 MM games are automatically also ladder.
The top50 page shows Ladder and Custom (instead of MM and Custom), this is even if a season is not ongoing (so that winners have a hall of fame of sorts).

As-is the lobby doesn't technically need anything. Any 1v1 MM game is automatically a ladder game. If the idea from https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/2850#issuecomment-940853724 is implemented there would be an additonal key in the JSON that describes possible queues for lobby to handle.

I believe this is "close to done" in that a migration should make it mergeable to test and working reasonably close to expectations. It doesn't implement the idea above that lets people opt out of ladder when playing MM. It doesn't keep track of historical seasons (but I think your approach with `if date > x then` would largely do that, if applied to the new rating). To generate a migration, as far as I can tell somebody just needs to open this project in Visual Studio, find a "generate migration" button and click it (which sounds trivial but I once managed to spend a day only to generate something that corrupted the database so I am scared).

Licho1:
was this project aborted @sprunk  ?

sprunk:
Yes. It was more of a crude proof-of-concept so not much loss.

sprunk:
This was an attempt at #2848 btw, that ticket contains more info

PR #2855, opened 2022-01-17
Fix likely typo (Add an 'e')

Sharkgamedev:
Unless this was explicitly left out for formatting reasons, fix the typo with a missing e in 'have'
And add a new line to be consistent with most other files in Infra.


sprunk:
image

PR #2879, opened 2022-05-07
Initial monokickstart for Steam

Porkch0p:
Add launcher from semi-portable install to Steam build. Improves compatibility for Linux environments.


Anarchid (approved):
- monokickstart binary seems to be missing +x perms and also fails to run on a mono-free ubuntu 20.04 with this message:
```
System.TypeInitializationException: The type initializer for 'System.Windows.Forms.XplatUI' threw an exception. ---> System.DllNotFoundException: libX11 assembly:<unknown assembly> type:<unknown type> member:(null)
```

However, @Licho1 's rebuilt mkbundled wrapper, which otherwise does **not** work on this mono-less system, starts launching properly. 

My understanding is that this adds the assemblies that were missed during the bundle step.

Porkch0p:
> * monokickstart binary seems to be missing +x perms

Changed in https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/2880/commits/daa0f5549def7708b4f197ca5e99819a546987e7

PR #2887, opened 2022-05-25
Add parser for Challonge export strings to tourney page

Aquanim:
My local git is misbehaving so I had to copy this code into the Github text editor, which may have mucked up the whitespace somehow. Merge with caution.


Aquanim:
I made a batch script to parse these strings. Still would be nice to have this on the server (saves a few button-clicks and the trouble of distributing the script to new tournament runners) but it's no longer urgent. I'll see about fixing this later, probably easiest to do it in a new PR.

PR #2903, opened 2022-06-26
GetResourceInfo for geteting map/mod info via wrapper

Licho1:
See https://github.com/ZeroK-RTS/Chobby/issues/829


Aquanim (approved):
Approving in case nobody more competent than myself shows up any time soon

PR #2914, opened 2022-08-10
Drop avatarwiki from the forum formatting link

sprunk:
Doesn't apply to mediawiki and avatarwiki is discouraged.


sprunk:
Doesn't apply to mediawiki and avatarwiki is discouraged.

Licho1 (approved):

PR #2929, opened 2023-01-04
Fix #2791 (game mode forum threads)

### LINKED ISSUE #2791: Game mode threads do not disappear.
No matter how much I click this button the thread does not disappear from this list. This was occurring before I posted in it. When I posted the item left the list, but since someone else posted the bug returned.

image

sprunk:
Was updating when downloading the JSON metadata instead of when actually displaying the thread.


Licho1 (approved):

Licho1:
Nice catch :-)

PR #2930, opened 2023-01-24
Blind commit to try to fix ladder rating.

GoogleFrog:
* Average WHR for MM games over the last three days rather than the last 30. Other ladders should be unaffected, and the last 30 days of MM activity are still displayed.
 * Attempt to replace the div0 into IsNan check with a more reasonable way of detecting activity, which makes the above possible.
 * Divide the amount that WHR leashing effects ladder rating by the number of people on the players team. It's about half an half in 1v1, down to a negligible change in large team games. In effect this scales the WHR effect to the classic elo effect, since elo change is divided by team mates.
 * Weighting between WHR leashing and classic elo now scales depending on difference, rather than being fixed at 50%.
 
 Completely untested.


Shared/PlasmaShared/GlobalConst.cs
```diff
@@ -121,15 +121,14 @@ static void SetMode(ModeType newMode)
         public const int MinDurationForPlanetwars = 0;
         public const int MaxDurationForPlanetwars = 60*60*3; // 3 hours
 
-        public static int LadderActivityDays => mode == ModeType.Live ? 30 : 90;
+        public static int LadderActivityDays => mode == ModeType.Live ? 3 : 90;
```
sprunk, line 124: Won't this mean that the ladder is now like, five people (whoever played in the last 3 days)?

sprunk:
> he variable for this is called "LadderActivityDays", but it is used in exactly one place, and that is where the average of the last X days is taken. If "LadderActivityDays" also control how long it takes for people to fall off the ladder, then we're dealing with some real spaghetti.

https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/34382be6e1a36e06adb169d5d223bba7d93bd556/ZkData/Ef/WHR/Player.cs#L249

This is the line where the constant is used. If it is low enough, a player may have zero games played. One of the next lines is then division by zero: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/34382be6e1a36e06adb169d5d223bba7d93bd556/ZkData/Ef/WHR/Player.cs#L259

This is fine and causes the value to be NaN. Elsewhere, at the end of this line, we've got `!float.IsNaN(p.avgElo)` which sets the `bool isRanked` field in the player reting struct: https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/master/ZkData/Ef/WHR/WholeHistoryRating.cs#L536

:spaghetti:

GoogleFrog:
That isn't fine at all! I assumed everyone had a rating defined on all days, even days they didn't play, since otherwise there would be a divide by 0. There was no check for div0 locally and letting one propagate out into the rest of the code, to be handled wherever this value is used, seemed like spaghetti. So... what even changing the rating average period would require duplicating the whole average elo system? I'm not good enough at blind pushing to do that.

sprunk:
By "fine" I just meant that division by zero is well enough defined not to crash or something, it becomes NaN reliably.


I think having a different timeout for fall-off-the-ladder wouldn't need to replicate the averaging system, it might basically boil down to a new number in the `Player` struct that gets calculated here like
```diff
+ lastDayPlayed = 0;
  for (int i = 0; i < days.Count; i++)
  {
+   // outside the `minDay` averaging check
+   if (days[i].weight > 0) // if any game played that day
+     lastDayPlayed = max(lastDayPlayed, days[i].day);
    if (days[i].day >= minDay)
    {
       weightSum += days[i].weight;
       varSum += days[i].weight * days[i].GetEloStdev() * days[i].GetEloStdev();
       avgElo += days[i].GetElo() * days[i].weight;
    }
 }
```
and then in the other line that sets ranked, replace `!isNaN(player.avgElo)` with something like `player.lastDayPlayed > (today - GlobalConst.PeriodToFallOffLadder)`.

Adding a new member to the structure involves a migration though so I can't blind push.

ZkData/Ef/WHR/Ranks.cs
```diff
@@ -120,14 +120,16 @@ public static float UpdateLadderRating(Account acc, RatingCategory cat, float ta
                 return (float)ladderElo;
             }
             var delta = targetRating - ladderElo;
-            delta *= GlobalConst.LadderEloSmoothingFactor; //smooth out elo changes
+            var classicWeight = 1 - Math.Abs(delta) / (30 + Math.Abs(delta));
+            delta *= GlobalConst.LadderEloSmoothingFactor; //smooth out elo changes.
```
DeinFreund, line 124: If you want to modify the rating delta, e.g. by dividing by the number of allies, please do so here, before the min/max clamping. Otherwise the change limit constants are no longer enforced.

DeinFreund:
> * Average WHR over the last three days rather than the last 30. 

The way the ladder rating was intended to work is that the ladder represents your performance within the ladder activity interval. So if the ladder activity interval is one month, then the ladder rating represents your average performance over the past month. Your ladder rating is only defined if you're active within that time frame; if you aren't, the ladder rating isn't defined and thus you're automatically off the ladder. This inherently guarantees that nobody can be on the ladder while not having any data within the ladder activity period.

This system was designed before the introduction of rolling-average smoothed ladder elo as we have it now. The rolling average should be enough to smooth out rating changes without having to also average over multiple activity days. Thus, if you want to reduce the averaging period, I would suggest to remove it entirely and use the most recent rating instead. Otherwise we're increasing complexity by maintaining two different parameters: Ladder activity days and ladder averaging period.

> * Divide the amount that WHR leashing effects ladder rating by the number of people on the players team. 

Note that WHR internally already assigns proportionally less weight to team games just how classic elo does, but that is unrelated to the rolling-average update formula. I don't see an issue with this change.

> * Weighting between WHR leashing and classic elo now scales depending on difference, rather than being fixed at 50%.

Is there some reasoning behind this weighting formula? It seems a bit weird to have high rating deviation swings being given to players that are close to their true rating. This seems to encourage an equilibrium where the rating is always a certain amount off from the true rating?

This is problematic for new players, where ladder elo is intentionally being started off at 1100, i.e. far below their true rating. With classic elo updates and only limited whr weighting, they get a small elo bonus every game. With this change they might rise in rating much faster and lose their elo bonus more quickly.

ZkData/Ef/WHR/Ranks.cs
GoogleFrog: Why? I ordered it like this intentionally. Why is it important that the WHR leash always pulls by at least 1? The constants never enforced a limit on the total change because classicEloChange doesn't interact with them.

DeinFreund: You're right, it must've been an oversight on my part. I think at least the allowGain and allowLoss parameters should also override the elo delta. Even though it gives the same result that is a cleaner way to implement it. 

Having a minimum of 1 is not really required since we now have the classic elo delta, it can be safely set to zero now.

GoogleFrog:
Multiplying classic elo and the leash rate by `sqrt(numPlayers)` sounds good to me.

Removing the rolling average sounds fine too. The code change required seems too large for me to be likely to do it though.

> Is there some reasoning behind this weighting formula? It seems a bit weird to have high rating deviation swings being given to players that are close to their true rating. This seems to encourage an equilibrium where the rating is always a certain amount off from the true rating?

Where do you get the high swings from? From my understanding, a victory in an even game gives each team a delta of 16 elo points, to spread among the team. So in an 8v8 each player would gain or lose 2. My goal is just to have a large region where ladder rating behaves "sanely", by which I mean it goes up and down in reasonable proportionate amounts. If a player plays five even games in an evening and wins three, then their ladder rating should be higher at the end.

Currently the extreme leashing means they might gain 1 point for each of the wins and lose 3 for each of the losses. This doesn't look sane to anyone who is tracking their ladder rating. Sure, tracking your rating this closely doesn't make mathematical sense. But people do it, and complain about what they see. The complaints seem fairly common.

What sort of equilibrium are you talking about precisely? I don't see how it will encourage ladder rating to always be, say, around 30 more than WHR. But it does seem likely that a low percentage of people will have ladder rating within 30 of WHR at any given time. I don't really care about a discrepancy this small though. Making ladder rating not annoy many people who track it seems much more important,

> This is problematic for new players, where ladder elo is intentionally being started off at 1100, i.e. far below their true rating. With classic elo updates and only limited whr weighting, they get a small elo bonus every game. With this change they might rise in rating much faster and lose their elo bonus more quickly.

I don't understand the problem here. The elo bonus is only for ladder rating, right? I assume nothing that happens in ladder rating flows back to mess with WHR. When new players are in this state, is it true that they will gain ladder rating for wins, and lose it for losses, and that the gains and losses won't differ by some large factor? If so, then I don't see a problem.

Licho1 (approved):

PR #2946, opened 2023-05-30
blind attempt to fix #2945

### LINKED ISSUE #2945: Replays through site aren't accessible (only for Linux systems?)
Screenshot_20230530_065015

I'm running the Steam Linux version natively, NOT through Proton.  Supposedly works for Windows users from what I've read.
Distro is Debian sid. 
Browser is Firefox.

Anarchid:
[x] compiles
[?] works


M5PR:
Maybe 

``url = EscapeDataString(urll);``

Is enough. Could also build a check whether it will work as a URI and if not, log it and handle it.

Anarchid:
Likely doesn't work

PR #2954, opened 2023-07-21
Updated readme to point to wiki

ghughes13:
If merged, this PR will:

1) Point developers to our development wiki. There is currently no info on contributing, build instructions, style guidelines, ect in out README & it makes since to have the location of that information easily accessible to developers. 

2) Adjusts heading syntax from alternative `===` to standard `#`. If we ever start adding more information to our README files and need heading levels past h2, we'd need to convert to the standard syntax anyway, so this is future proofing/planning-ish.


sprunk (approved):

sprunk:
I ran some check which seems to be required for merge. It failed for some reason that looks unrelated. The button to re-run it is gone and I'm not sure what now.

sprunk (approved):

PR #2968, opened 2023-12-23
Map choices for 4-way map poll

sprunk:
* no longer reserves 2 spots for maps from the top of the map ladder (avoids a feedback loop, among other things)
* always suggests at least Featured maps (can still set lower rarity maps manually in private rooms)


Licho1 (commented):
Let's change it to have ratio configurable on website (there is some settings feature for this)

Licho1 (approved):

PR #2980, opened 2024-04-07
Modoptions: send current date

sprunk:
Lets you have events (e.g. Easter) based on time.

### DIFF
--- a/ZkLobbyServer/SpringieInterface/StartSetup.cs
+++ b/ZkLobbyServer/SpringieInterface/StartSetup.cs
@@ -292,6 +292,13 @@ namespace ZeroKWeb.SpringieInterface
                  * See https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/master/Shared/LobbyClient/DedicatedServer.cs#L317 */
                 ret.ModOptions["sendSpringieData"] = "1";
 
+                /* Current date. Synced Lua code cannot access `os.date`
+                 * which is needed for gameside events (Easter etc) */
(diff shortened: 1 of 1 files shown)


ZkLobbyServer/SpringieInterface/StartSetup.cs
```diff
@@ -292,6 +292,13 @@ public static LobbyHostingContext GetDedicatedServerStartSetup(LobbyHostingConte
                  * See https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/master/Shared/LobbyClient/DedicatedServer.cs#L317 */
                 ret.ModOptions["sendSpringieData"] = "1";
 
+                /* Current date. Synced Lua code cannot access `os.date`
+                 * which is needed for gameside events (Easter etc) */
+                var date = DateTime.UtcNow;
+                ret.ModOptions["date_day"] = date.Day.ToString();
+                ret.ModOptions["date_month"] = date.Month.ToString();
+                ret.ModOptions["date_year"] = date.Year.ToString();
+
```
Licho1, line 301: any reason why is it split to multiple keys and not just as a combined string?

Licho1 (approved):

ZkLobbyServer/SpringieInterface/StartSetup.cs
sprunk: Ingame would want to split it into individual numbers anyway to get logic like
```lua
local xmas = month == 12 and day > 20
```

PR #2985, opened 2024-05-05
Admins can see everybody in lobby

sprunk:


ZkLobbyServer/ZkLobbyServer.cs
```diff
@@ -212,6 +212,9 @@ public bool CanUserSee(ConnectedUser uWatcher, ConnectedUser uWatched)
             // admins always visible
             if (uWatched.User?.IsAdmin == true) return true;
 
+            // admins see everybody
+            if (uWatcher.User?.IsAdmin == true) return true;
```
Licho1, line 216: not sure about this, can it lead to some confusion? Perhaps we should just increase limits?

sprunk: The problem with the current state of affairs, where admins are subject to the same "only see 50 random people" rule as everybody else, is that sometimes you need to send a message to some random player who needs help, or join his game, and he may be outside of those 50.

Increasing the limit is probably ok too, but there can always be a situation where we get more players and reach the limit again. Admins already have an exception where all players always see them so making it work both ways sounds good too.

Licho1 (approved):

PR #3020, opened 2025-07-30
Prevent players in waitlist from being considered in votes #2780

ykc1n:
`Battlecommand.GetRunPermissions()` will now check if the player is in the waitlist, before considering the player as part of a vote.

 I added a function in `ServerBattle`, `public bool ServerBattle.IsInWaitlist(username)`, that takes in a player's username, and returns a boolean to check whether or not they are in the waitlist.

The logic in this function is basically identical to the first few lines in `ServerBattle.StartGame()` . As far as I can tell, this is how you determine if a player is in the waitlist. 

If a player is considered part of the waitlist, they will not be able to vote, and the bot will specify that players in the waitlist should not be able to vote.

Some thoughts / concerns:
- Not sure if there should be any difference between spectator and waitlisted specified in the messages.

- Not sure if im missing any logic or edge cases in `IsInWaitlist()`. It seems to be fine, since its exactly the code that determines what players actually play or not when a game is started, but still. 

- Not sure if its safe to call `battle.IsInWaitlist()`; I dont know if battle can be null when a vote is caused, and if that will cause some crash.

### DIFF
--- a/ZkLobbyServer/ServerBattle.cs
+++ b/ZkLobbyServer/ServerBattle.cs
@@ -84,7 +84,7 @@ namespace ZkLobbyServer
         public bool IsDefaultGame { get; private set; } = true;
         public bool IsCbalEnabled { get; private set; } = true;
 
-        public bool TimeQueueEnabled = true;//> DynamicConfig.Instance.TimeQueueEnabled && true;//(Mode == AutohostMode.Teams || Mode == AutohostMode.Game1v1 || Mode == AutohostMode.GameFFA);
+        public override bool TimeQueueEnabled => DynamicConfig.Instance.TimeQueueEnabled && (Mode == AutohostMode.Teams || Mode == AutohostMode.Game1v1 || Mode == AutohostMode.GameFFA);
 
         public MapSupportLevel MinimalMapSupportLevelAutohost { get; protected set; } = MapSupportLevel.Featured;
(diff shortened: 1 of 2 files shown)


GoogleFrog:
Where is the part where the votes required is set to only take the number of non-waitlist players into account? Or is that already the case?

ykc1n:
It should already be the case:

https://github.com/ykc1n/Zero-K-Infrastructure/blob/d038d39ac03c15bce40758391ec97e80790b22d6/ZkLobbyServer/ServerBattle.cs#L608C1-L638C1

numvoters only considers users who are not given an ineligibilty reason. 


https://github.com/ykc1n/Zero-K-Infrastructure/blob/d038d39ac03c15bce40758391ec97e80790b22d6/ZkLobbyServer/autohost/Commands/BattleCommand.cs#L61C2-L71C1

GetIneligibiltyReason simply just uses GetRunPermissions, which is the function i changed to reject a player who is in the waitlist.

Stuart1998:
Are there any blockers preventing this from being merged?

sprunk:
It could use testing (e.g. for the concerns in the first post like "can a battle be null?"), and possibly a @Licho1 review (e.g. maybe this is performant enough for gamestart, but not for votes which tend to get spammed?).

PR #3067, opened 2026-09-23
Auto-noelo chickens maps

Stuart1998:
Maps tagged as chickens are treated the same as maps with the special tag for rating purposes. Not sure how to test this to make sure that it works?


Zero-K.info/AppCode/LadderCalculator.cs
```diff
@@ -45,7 +45,7 @@ private static List<AwardItem> CalculateAwards(ZkDataContext db)
                 db.SpringBattles.Where(
                         x =>
                             (x.StartTime >= monthStart) && (x.HasBots == false) &&
-                            (x.ResourceByMapResourceID.MapSupportLevel >= MapSupportLevel.Supported) && (x.ResourceByMapResourceID.MapIsSpecial == false))
+                            (x.ResourceByMapResourceID.MapSupportLevel >= MapSupportLevel.Supported) && (x.ResourceByMapResourceID.MapIsSpecial == false) && (x.ResourceByMapResourceID.MapIsChickens == false))
```
sprunk: awards don't really do anything and there are some chicken-specific ones, should probably be kept

sprunk:
> Not sure how to test this to make sure that it works?

IIRC on the zk mediawiki there are instructions on how to setup a local server, you'd then have to connect to it with two accounts and play a game.

Alternatively this looks pretty safe so we could also just merge and test it on the ZK testserver.

